PR #18, opened 2022-03-09
summit attendees new filter criteria by has virtual checkin

romanetar:
ref https://tipit.avaza.com/project/view/295592#!tab=task-pane&task=2854051


smarcet (approved):
Looks good to me

PR #20, opened 2022-03-15
remove overlapping validation rule when create or update a selection plan

romanetar:
ref https://tipit.avaza.com/project/view/304272#!tab=task-pane&task=2860688


smarcet (approved):
looks good to me

PR #24, opened 2022-03-28
TRACK_GROUPS schedule filter label tweak

romanetar:
ref https://tipit.avaza.com/project/view#!tab=task-pane&groupby=MyTaskDueDate&view=vertical&task=2870570&fileview=grid

### DIFF
--- a/database/migrations/model/Version20220328170502.php
+++ b/database/migrations/model/Version20220328170502.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Database\Migrations\Model;
+
+use Doctrine\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema as Schema;
+
+class Version20220328170502 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema): void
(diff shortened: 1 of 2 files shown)


database/migrations/model/Version20220328170502.php
```diff
@@ -0,0 +1,28 @@
...
+
+class Version20220328170502 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema): void
+    {
+        $sql = <<<SQL
+UPDATE `SummitScheduleFilterElementConfig` SET Label = 'Category Groups' WHERE Type = 'TRACK_GROUPS';
```
smarcet, line 16: this is not rite u should only update it if the Former label value was 'Categories Groups' otherwise u will overwrite legit customizations
regards

smarcet (changes requested):

database/migrations/model/Version20220328170502.php
romanetar: Done

smarcet (approved):
looks good to me

PR #36, opened 2022-04-26
Validate if the company to add to the summit is already associated

romanetar:
ref https://tipit.avaza.com/project/view/304272#!tab=task-pane&task=2896768


smarcet (approved):
LGTM

PR #62, opened 2022-08-01
fix company model & add ingest test case

romanetar:
ref https://tipit.avaza.com/project/view/304272#!tab=task-pane&task=2939128


smarcet (approved):
LGTM

PR #78, opened 2022-09-23
add new filter by ticket types to registration invitations

romanetar:
ref https://tipit.avaza.com/project/view/203215#!tab=task-pane&task=3045398 


smarcet (approved):
LGTM

PR #96, opened 2022-11-18
Feature/presentation action types global removal check

romanetar:
Validate if action type is attached to any selection plan before remove it globally


smarcet (approved):
LGTM

PR #118, opened 2023-02-07
Audit log formatter fix: object to string conversion error

romanetar:
Fix for reported error by Slack


smarcet (approved):
LGTM

PR #134, opened 2023-04-19
bypass for a period of time subsequent invocations on connection error to Swift

romanetar:
ref https://tipit.avaza.com/project/view#!tab=task-pane&groupby=MyTaskDueDate&view=vertical&task=3262392&stafilter=NotComplete&fileview=grid

### DIFF
--- a/app/Services/FileSystem/Swift/SwiftServiceProvider.php
+++ b/app/Services/FileSystem/Swift/SwiftServiceProvider.php
@@ -12,6 +12,8 @@
  * limitations under the License.
  **/
 
+use Carbon\Carbon;
+use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Log;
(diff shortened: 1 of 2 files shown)


app/Services/FileSystem/Swift/SwiftServiceProvider.php
```diff
@@ -43,6 +74,8 @@ public function boot()
         Storage::extend('swift', function ($app, $config) {
 
             try {
```
smarcet, line 76: let adds a return null if the config object has not set the auth url the region or the container

smarcet (changes requested):

app/Services/FileSystem/Swift/SwiftServiceProvider.php
romanetar: done

smarcet (approved):
LGTM

PR #197, opened 2023-09-27
filter speakers and submitters by presentation media uploads

romanetar:
ref https://tipit.avaza.com/project/view#!tab=task-pane&groupby=MyTaskDueDate&view=vertical&task=3385040&stafilter=NotComplete&fileview=grid

### DIFF
--- a/app/Http/Utils/Filters/Filter.php
+++ b/app/Http/Utils/Filters/Filter.php
@@ -634,4 +634,22 @@ final class Filter
     {
         return $this->originalExp;
     }
+
+    /**
+     * @param FilterElement[] $e
+     * @return array
+     */
+    public static function getValueFromFilterElement(array $e): array {
+        $v = [];
+        foreach($e as $f){
+            if(is_array($f->getValue())){
+                foreach ($f->getValue() as $iv){
+                    $v[] = $iv;
+                }
+            }
+            else
+                $v[] = $f->getValue();
+        }
+        return $v;
+    }
 }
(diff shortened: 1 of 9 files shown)


app/Http/Utils/Filters/Filter.php
```diff
@@ -634,4 +634,22 @@ public function getOriginalExp()
     {
         return $this->originalExp;
     }
+
+    /**
+     * @param FilterElement[] $e
+     * @return array
+     */
+    public static function getValueFromFilterElement(array $e): array {
```
smarcet, line 642: i am no so sure that this should be a static method
IMHO should be an instance method using https://github.com/OpenStackweb/summit-api/blob/fcac0ee396cc8dbf287deef825f103ed04d29107/app/Http/Utils/Filters/Filter.php#L111

romanetar: done

app/Models/Foundation/Summit/Speakers/PresentationSpeaker.php
```diff
@@ -884,27 +884,27 @@ public function getAcceptedPresentations
 
         if (!is_null($filter)) {
             if ($filter->hasFilter("presentations_selection_plan_id")) {
-                $v = Filter::getValueFromFilterElement($filter->getFilter("presentations_selection_plan_id"));
+                $v = $filter->getValueFromFilterElement($filter->getFilter("presentations_selection_plan_id"));
```
smarcet, line 887: use $this->getFilter inside getValueFromFilterElement to avoid this double call

smarcet (changes requested):

smarcet (approved):
LGTM

PR #255, opened 2024-05-10
feat: add support for %NULL% and %EMPTY% filter tokens

romanetar:
ref https://tipit.avaza.com/project/view#!tab=task-pane&groupby=MyTaskDueDate&view=vertical&task=3544153&stafilter=NotComplete&fileview=grid

ref https://tipit.avaza.com/project/view#!tab=task-pane&groupby=MyTaskDueDate&view=vertical&task=3544163&stafilter=NotComplete&fileview=grid

### DIFF
--- a/tests/FilterParserTest.php
+++ b/tests/FilterParserTest.php
@@ -805,4 +805,82 @@ DQL;
             'email_sent' => ['sometimes', new Boolean()],
         ]);
     }
+
+    public function testApplyCompanyFilterEMPTYtoAttendeesList(){
+        $filters_input = [
+            'company==%EMPTY%'
+        ];
+
+        $filter = FilterParser::parse($filters_input, [
+            'company' => ['==']
+        ]);
+
+        $em = Registry::getManager(SilverstripeBaseModel::EntityManager);
+        $query = new QueryBuilder($em);
+        $query->select("e")
+            ->from(\models\summit\SummitAttendee::class, "e")
+            ->leftJoin('e.summit', 's')
+            ->leftJoin('e.member', 'm')
+            ->leftJoin('e.tickets', 't');
+
+        $filter->apply2Query($query, [
+            'company' => new DoctrineFilterMapping("e.company_name :operator :value")
+        ]);
+
+        $dql = $query->getDQL();
+
+        $this->assertNotEmpty($dql);
+
+        $params = $query->getParameters();
+
+        $this->assertEquals(1, $params->count());
+        $this->assertEmpty($params[0]->getValue());
+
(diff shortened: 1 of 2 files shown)


smarcet:
@romanetar i do think that there is a simpler approach at FilterElement class

```
static $symbols = ['%NULL%' => 'NULL','%EMPTY%' => ''];
public static function mapValueSymbols(string $val):string{
    if(isset(self::$symbols[strtoupper($val)])) return self::$symbols[strtoupper($val)];
    return $val;
}
/**
 * @return mixed
 */
public function getValue()
{
    $value = $this->value;;
    if(is_array($value)){
        $res = [];
        foreach ($value as $val){
               $res[]= empty($val) ? '' : self::mapValueSymbols($val);
        }
        $value = $res;
    }
    else {
        $value = self::mapValueSymbols($value);
    }

    switch($this->operator)
    {
        case 'like':
            if(is_array($value)){
                $res = [];
                foreach ($value as $val){
                    $res[]= empty($val) ? '' : "%".$val."%";
                }
                return $res;
            }
            return empty($value) ? '' : "%".$value."%";
            break;
        case 'start_like':
            if(is_array($value)){
                $res = [];
                foreach ($value as $val){
                    $res[]= empty($val) ? '' : $val."%";
                }
                return $res;
            }
            return  empty($value) ? '' : $value."%";
            break;
        default:
            return $value;
            break;
    }
}
```

smarcet (changes requested):

tests/FilterParserTest.php
```diff
@@ -805,4 +805,37 @@ public function testDuplicateCriteria(){
             'email_sent' => ['sometimes', new Boolean()],
         ]);
     }
+
+    public function testApplyCompanyFilterEMPTYtoAttendeesList(){
+        $filters_input = [
+            'company==%EMPTY%'
```
smarcet, line 811: please create a test with OR
 'company==%EMPTY%||1||2'

tests/FilterParserTest.php
```diff
@@ -805,4 +805,37 @@ public function testDuplicateCriteria(){
...
+        $filter->apply2Query($query, [
+            'company' => new DoctrineFilterMapping("e.company_name :operator :value")
+        ]);
+
+        $dql = $query->getDQL();
+
+        $this->assertNotEmpty($dql);
+
+        //if DQL is not well-formed this will fail
+        $sql = $query->getQuery()->getSQL();
```
smarcet, line 835: please check params

smarcet (changes requested):

smarcet (approved):
LGTM

PR #277, opened 2024-07-01
feat: inject default email template vars

romanetar:
ref https://tipit.avaza.com/project/view/203215#!tab=task-pane&task=3589655

### DIFF
--- a/app/Jobs/Emails/AbstractSummitEmailJob.php
+++ b/app/Jobs/Emails/AbstractSummitEmailJob.php
@@ -13,6 +13,7 @@
  **/
 
 use Illuminate\Support\Facades\App;
+use Illuminate\Support\Facades\Config;
 use Illuminate\Support\Facades\Log;
 use models\summit\ISummitRepository;
 use models\summit\Summit;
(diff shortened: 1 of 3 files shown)


config/marketing.php
```diff
@@ -14,5 +14,14 @@
 
 return [
     'base_url' => env('MARKETING_API_BASE_URL', null),
-    'cache_ttl' => env('MARKETING_VARS_CACHE_TTL', 60)
+    'cache_ttl' => env('MARKETING_VARS_CACHE_TTL', 60),
+
+    'default_email_template_vars' => [
```
smarcet, line 19: use default values here on env ( )

app/Jobs/Emails/AbstractSummitEmailJob.php
```diff
@@ -75,8 +76,22 @@ public function __construct
         $payload[IMailTemplatesConstants::registration_link] = $summit->getRegistrationLink();
         $payload[IMailTemplatesConstants::virtual_event_site_link] = $summit->getVirtualSiteUrl();
 
+        $payload = array_merge($payload, self::getMarketingVariables($summit));
+        parent::__construct($payload, $template_identifier, $to_email, $subject, $cc_email, $bcc_email);
+    }
+
+    private static function getMarketingVariables(Summit $summit) {
         $marketing_api = App::make(IMarketingApi::class);
         $marketing_vars = $marketing_api->getConfigValues($summit->getId(), 'EMAIL_TEMPLATE_');
```
smarcet, line 85: check that marketing_api is not null

app/Jobs/Emails/AbstractSummitEmailJob.php
```diff
@@ -75,8 +76,22 @@ public function __construct
...
+
+        $default_email_template_vars = Config::get('marketing.default_email_template_vars');
+        foreach ($default_email_template_vars as $key => $value) {
+            $key = strtoupper($key);
+            if (!array_key_exists($key, $marketing_vars)) {
```
smarcet, line 90: if not just exists
u need to check that also has something, if not set
then set the default one

app/Jobs/Emails/AbstractSummitEmailJob.php
smarcet, line 85: use a try catch here
and on the catch set the marketing_vars = []

PR #302, opened 2025-02-24
feat: check if current_member is null before checking if is authorized

romanetar:
ref https://tipit.avaza.com/project/view#!tab=task-pane&groupby=MyTaskDueDate&view=vertical&task=3738274&stafilter=NotComplete&fileview=grid


smarcet (approved):
LGTM

PR #332, opened 2025-07-10
feat: update add badge scan logic to accept encrypted qr code

smarcet:
ref https://tipit.avaza.com/project/view/383698#!tab=task-pane&task=3817804


romanetar (approved):
LGTM

PR #357, opened 2025-09-30
Feature | Extend Swagger Coverage for controller `CountriesApiController`

matiasperrone-exo:
## Task:
Ref: https://app.clickup.com/t/86b6rgvu3

### DIFF
--- a/app/Swagger/schemas.php
+++ b/app/Swagger/schemas.php
@@ -350,13 +350,3 @@ class RSVPUpdateRequestSchema_{
     ]
 )]
 class RSVPAdminAddRequestSchema {}
-
-#[OA\Schema(
-    schema: 'ISOElementSchema',
-    type: 'object',
-    properties: [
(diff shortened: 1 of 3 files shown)


smarcet (changes requested):
@matiasperrone-exo please review

app/Swagger/schemas.php
```diff
@@ -348,4 +349,33 @@ class RSVPUpdateRequestSchema_{
 
     ]
 )]
-class RSVPAdminAddRequestSchema {}
\ No newline at end of file
+class RSVPAdminAddRequestSchema {}
+
+#[OA\Schema(
```
smarcet, line 354: @matiasperrone-exo we need to start to break down this schema file on several files ( like one file responsability
* marketplace
* rsvp
* general 
and so

smarcet (changes requested):
@matiasperrone-exo please review comments

PR #366, opened 2025-09-30
Feature | Extend Swagger Coverage for controller Apis/Protected/Main/OAuth2ChunkedFilesApiController.php

matiasperrone-exo:
- Add controller's response to OpenAPI schema


caseylocker (changes requested):
There's a status property for uploads that seems to be missing in schemas.php.
After the 'done' property there should also be a 'status' property:

new OA\Property(property: 'status', type: 'boolean', example: true, description: 'Upload status indicator'),

Minor but when that's fixed it should be good to go.

app/Http/Controllers/Apis/Protected/Main/OAuth2ChunkedFilesApiController.php
matiasperrone-exo: @caseylocker As you can see there is only "done" as field

PR #375, opened 2025-10-02
Feature | Extend Swagger Coverage for controller Apis/Protected/Summit/OAuth2SummitMediaFileTypeApiController.php

matiasperrone-exo:


caseylocker (changes requested):
@matiasperrone-exo looks like this one also relies on 'summit_rsvp_oauth2' being a defined security schema but it doesn't exist in the Pr in app/Swagger/schemas.php
It looks like once that's actually defined it will fix at least a few of these PRs.

matiasperrone-exo:
> @matiasperrone-exo looks like this one also relies on 'summit_rsvp_oauth2' being a defined security schema but it doesn't exist in the Pr in app/Swagger/schemas.php It looks like once that's actually defined it will fix at least a few of these PRs.

@caseylocker thanks the issue is now fixed and the docs are compiling now.

caseylocker (changes requested):
Change the "tags" for the GET all to match the other ones - tags: ['Summit Media File Types'],

None of the 5 endpoints have operationId defined. Add for API client generation:
GET all: operationId: 'getAllSummitMediaFileTypes'
GET one: operationId: 'getSummitMediaFileType'
POST: operationId: 'createSummitMediaFileType'
PUT: operationId: 'updateSummitMediaFileType'
DELETE: operationId: 'deleteSummitMediaFileType'

Missing type: 'object' in Paginated Schema
Should be:
new OA\Schema(
    type: 'object', 
    properties: [

PR #379, opened 2025-10-02
Feature | Extend Swagger Coverage for controller ReleasesApiController

andrestejerina97:
ref: https://app.clickup.com/t/86b6wkh2w


caseylocker (changes requested):
The API Path is wrong. "path: '/api/v1/releases/current'," should be "path: '/api/public/v1/releases/current',"

The "anyOf" usage is incorrect. 
The "type: 'SoftwareComponent'" isn't valid and to include it you'd need a ref to a schema. Just include the integer array and be more explicit in the description...
```
properties: [
    // ... existing properties ...
    new OA\Property(
        property: 'components',
        type: 'array',
        items: new OA\Items(type: 'integer'),
        description: 'Array of component IDs. Use ?expand=components for full objects'
    ),
]
// Remove anyOf entirely
```

PR #394, opened 2025-10-09
Feature | Extend Swagger Coverage for controller `OAuth2SummitTaxTypeApiController`

matiasperrone-exo:
## Task:
Ref: https://app.clickup.com/t/86b6wkh40


caseylocker (changes requested):
@matiasperrone-exo Doc generation is broken here like in pr 393.
Please verify existence of, or create
new OA\Parameter(ref: '#/components/parameters/page'),
new OA\Parameter(ref: '#/components/parameters/per_page'),

Also verify 'OAuth2' sec schema. 

For 'SummitTaxType' look at the 'ticket_types' property. 
Without relations=ticket_types: Field not present
With relations=ticket_types: Returns array of integers (IDs): [1, 2, 3]
With expand=ticket_types: Returns array of full objects
Should probably be something like:
new OA\Property(
    property: 'ticket_types',
    type: 'array',
    nullable: true,
    items: new OA\Items(type: 'integer'),
    example: [1, 2, 3],
    description: 'Array of ticket type IDs (only present when relations=ticket_types). When expanded with expand=ticket_types, contains full ticket type objects instead of IDs.'
),

caseylocker (approved):
Approved

PR #402, opened 2025-10-14
Feature | Extend Swagger Coverage for controller `OAuth2SummitScheduleSettingsApiController`

matiasperrone-exo:
## Task:

Ref: https://app.clickup.com/t/86b6wkh7d

### DIFF
--- a/app/Swagger/SummitSchemas.php
+++ b/app/Swagger/SummitSchemas.php
@@ -3,6 +3,36 @@
 namespace App\Swagger\schemas;
 
 use OpenApi\Attributes as OA;
+#[OA\Schema(
+    schema: 'SummitScheduleConfigContent',
+    type: 'object',
+    properties: [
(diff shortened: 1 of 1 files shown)


caseylocker (changes requested):
@matiasperrone-exo only one issue with the namespace. 

Wrong Namespace in Security Schema
app/Swagger/Security/SummitScheduleSettingsAuthSchema.php
namespace App\Swagger\Security;  //  WRONG
Should be:
namespace App\Swagger\schemas;  // CORRECT

While you're in there is there any reason that you're using numerical responses missed with constants in responses? E.g.
responses: [
            new OA\Response(response: 204, description: 'No Content'), // <- Why?
            new OA\Response(response: Response::HTTP_UNAUTHORIZED, description: "Unauthorized"),

We should stick with one or the other.

caseylocker (approved):
Looks good. @smarcet if you agree please merge.

PR #409, opened 2025-10-14
Feature | Extend Swagger Coverage for controller `OAuth2SummitDocumentsApiController`

matiasperrone-exo:
## Task:

Ref: https://app.clickup.com/t/86b6wkh83

### DIFF
--- a/app/Swagger/Security/SummitAuthSchema.php
+++ b/app/Swagger/Security/SummitAuthSchema.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace App\Swagger\Security;
+namespace App\Swagger\schemas;
 
 use App\Security\SummitScopes;
 use OpenApi\Attributes as OA;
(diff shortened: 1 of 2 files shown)


caseylocker (changes requested):
@matiasperrone-exo name space issue...
Wrong Namespace in Security Schema
app/Swagger/Security/SummitAuthSchema.php
namespace App\Swagger\Security;  //  WRONG
Should be:
namespace App\Swagger\schemas;  //  CORRECT


Wrong x: Parameter Key Name
Write endpoints use 'authz_groups' instead of the standard 'required-groups':
x: [
    'authz_groups' => [  //  Wrong key name
        IGroup::SuperAdmins,
        ...
    ]
],
Should be:
x: [
    'required-groups' => [  //  Correct key name
        IGroup::SuperAdmins,
        ...
    ]
],
This affects 7 endpoints (all write operations).

smarcet (approved):
LGTM

PR #419, opened 2025-10-17
Feature | Extend Swagger Coverage for controller `OAuth2ElectionsApiController`

andrestejerina97:
ref: https://app.clickup.com/t/86b6wkh9k


caseylocker (changes requested):
Incorrect namespace as per project standards for Swagger
App\Swagger\security should be App\Swagger\schemas

Public  endpoints have the wrong path.
Wrong:
/api/v1/elections/current	
Right:
/api/public/v1/elections/current

Wrong
/api/v1/elections/current/candidates
Right
/api/public/v1/elections/current/candidates

Wrong
/api/v1/elections/current/candidates/gold
Right
/api/public/v1/elections/current/candidates/gold

PR #422, opened 2025-10-20
Feature | Extend Swagger Coverage for controller `OAuth2SummitProposedScheduleApiController`

matiasperrone-exo:
## Task:
Ref: https://app.clickup.com/t/86b6wkha7


caseylocker (changes requested):
@matiasperrone-exo 

As per our decision to limit sub schemas to keep this as simple as possible, and since the use of anyOf is incorrect here...
Remove the anyOf blocks and keep only the _id properties with descriptions noting expansion behavior.

caseylocker (approved):
This is approved.

PR #425, opened 2025-10-20
Feature: Extend Swagger Coverage for controller OAuth2SummitsEventTypesApiController

andrestejerina97:
ref: https://app.clickup.com/t/86b6wkha7

### DIFF
--- a/app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitsEventTypesApiController.php
+++ b/app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitsEventTypesApiController.php
@@ -896,13 +896,6 @@ final class OAuth2SummitsEventTypesApiController extends OAuth2ProtectedControll
         ],
         parameters: [
             new OA\Parameter(
-                name: "access_token",
-                in: "query",
-                required: false,
-                description: "OAuth2 access token (alternative to Authorization: Bearer)",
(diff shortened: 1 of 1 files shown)


smarcet (changes requested):
@andrestejerina97 please review comments

caseylocker (changes requested):
@matiasperrone-exo The POST in app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitsEventTypesApiController.php should be a PUT based on the route.

Incorrect
```
#[OA\Post(
    path: "/api/v1/summits/{id}/event-types/{event_type_id}/summit-documents/{document_id}",
    operationId: "addSummitDocument",
```

Correct
```
#[OA\Put(
    path: "/api/v1/summits/{id}/event-types/{event_type_id}/summit-documents/{document_id}",
    operationId: "addSummitDocument",
```

Is app/Swagger/Models/SummitSchema.php complete?
There is a TODO comment in there.

caseylocker (changes requested):
PUT operations document HTTP 200 but updated() actually returns a 201.
updateEventTypeBySummit, addSummitDocument, removeSummitDocument Change Response::HTTP_OK to Response::HTTP_CREATED

Missing security attribute on protected GET endpoint in getEventTypeBySummit.
Thanks for adding the public endpoint but the protected one needs documentation of the security attribute.

PR #431, opened 2025-10-22
Feature | Extend Swagger Coverage for controller `OAuth2PresentationCategoryGroupController`

matiasperrone-exo:
## Task:
Ref: https://app.clickup.com/t/86b6wkhap


caseylocker (changes requested):
All endpoints use:
`tags: ['track-groups']`
Should be Title Case:
`tags: ['Track Groups']`

Invalid Schema Reference
new OA\Schema(type: 'PresentationCategory')  // INVALID

Needs a Security Schema with appropriate scopes instead of Bearer...

PR #432, opened 2025-10-22
feat: Extend Swagger Coverage for controller OAuth2SummitOrderExtraQuestionTypeApiController

andrestejerina97:
ref: https://app.clickup.com/t/86b6wkhbh


caseylocker (changes requested):
All endpoints lack security attributes even though they're OAUTH endpoints.
They should include the appropriate scopes.

PR #498, opened 2026-02-11
Feature: add a new event for many-to-many relationships

andrestejerina97:
ref: https://app.clickup.com/t/86b84w9x2

## Summary by CodeRabbit

* **New Features**
  * Audit logs now track many-to-many relationship changes with detailed information about which entities were added or removed.
  * PresentationCategoryGroup entity modifications are now audited.
  * Relationship changes display specific entity IDs for improved traceability and visibility.


andrestejerina97:
@martinquiroga-exo ready to review

martinquiroga-exo (approved):
LGTM

caseylocker (commented):
I'm going to defer to whatever Seb comes up with but this is what I'm seeing. 

The auditCollection() method changed the calling convention for _all collection updates_  not just ManyToMany. For non-ManyToMany collections, it now passes $owner (the entity) as the subject  instead of $col (the collection itself), and $payload instead of [] as the change set.

Both strategies downstream expect a PersistentCollection as $subject for EVENT_COLLECTION_UPDATE:
                                          
- OTLP strategy: instanceof PersistentCollection checks fail, formatter returns null, audit entries silently dropped
- DB strategy: calls count($subject) and $subject->getSnapshot() on the entity — throws \Error (not \Exception), uncaught by the catch block

Fix: Unless the intent was to actually change the behavior for all for some reason, only change the subject/payload for ManyToMany events; preserve the original $strategy->audit($col, [], EVENT_COLLECTION_UPDATE, $ctx) call for non-ManyToMany collections.

andrestejerina97:
@caseylocker it's ready to review

caseylocker (changes requested):
Two fixes needed before merge:

1. app/Audit/AuditLogOtlpStrategy.php — Add M2M event mappings

mapEventTypeToAction() and getLogMessage() need explicit cases for EVENT_COLLECTION_MANYTOMANY_UPDATE (→ ACTION_COLLECTION_UPDATE / LOG_MESSAGE_COLLECTION_UPDATED) and EVENT_COLLECTION_MANYTOMANY_DELETE (→ ACTION_DELETE / LOG_MESSAGE_DELETED). Without this, M2M audit entries hit Elasticsearch with action=unknown, making them invisible to dashboards/queries filtering on action type.

buildAuditLogData() should also extract collection metadata from $change_set['collection'] for M2M events, since the subject is the owner entity (not a PersistentCollection).

2. app/Audit/AuditLogFormatterFactory.php — Fix generic M2M fallback

Lines 110-123: when no context-specific formatter exists, the fallback creates EntityCollectionUpdateAuditLogFormatter and passes the owner entity as $subject. That formatter calls $subject->getInsertDiff() (line 50 of EntityCollectionUpdateAuditLogFormatter), which doesn't exist on the owner entity. This either silently drops the audit entry or throws an uncaught \Error. The fallback should pass $change_set['collection'] as the subject instead of the owner, or use a M2M-aware generic formatter.

Why this blocks merge: Without fix 1, OTLP audit data is misclassified. Without fix 2, any M2M collection change on an entity without a registered custom formatter will silently fail or crash.

app/Audit/ConcreteFormatters/EntityManyToManyCollectionDeleteAuditLogFormatter.php
```diff
@@ -0,0 +1,85 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+/**
+ * Formatter for Many-to-Many collection deletions
+ */
+class EntityManyToManyCollectionDeleteAuditLogFormatter extends AbstractAuditLogFormatter
```
smarcet: @andrestejerina97 rename it to DefaultEntityManyToManyCollectionDeleteAuditLogFormatter
that communicates intention

app/Audit/ConcreteFormatters/EntityManyToManyCollectionUpdateAuditLogFormatter.php
```diff
@@ -0,0 +1,91 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+/**
+ * Formatter for Many-to-Many collection updates
+ */
+class EntityManyToManyCollectionUpdateAuditLogFormatter extends AbstractAuditLogFormatter
```
smarcet: @andrestejerina97 @andrestejerina97 rename it to DefaultEntityManyToManyCollectionUpdateAuditLogFormatter
that communicates intention

andrestejerina97: Done. Renamed to DefaultEntityManyToManyCollectionDeleteAuditLogFormatter.

martinquiroga-exo (changes requested):
Current changes does not adhere to the proposed ADR.

martinquiroga-exo (approved):
LGTM

caseylocker (approved):
LGTM codewise though there are merge conflicts currently.

smarcet (changes requested):
@andrestejerina97 @martinquiroga-exo @caseylocker please review comments

smarcet (approved):
LGTM

PR #499, opened 2026-02-18
feat: add new formatters for signs and settings

andrestejerina97:
ref: https://app.clickup.com/t/86b8jav0q

### DIFF
--- a/app/Audit/ConcreteFormatters/SummitDocumentAuditLogFormatter.php
+++ b/app/Audit/ConcreteFormatters/SummitDocumentAuditLogFormatter.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace App\Audit\ConcreteFormatters;
+
+/**
+ * Copyright 2026 OpenStack Foundation
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+use App\Audit\AbstractAuditLogFormatter;
+use App\Audit\Interfaces\IAuditStrategy;
+use models\summit\SummitDocument;
+use Illuminate\Support\Facades\Log;
+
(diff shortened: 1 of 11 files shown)


martinquiroga-exo (approved):
LGTM

caseylocker (changes requested):
SummitScheduleConfigAuditLogFormatter.php:36 calls getIsDefault() which doesn't exist on SummitScheduleConfig.

The entity only has isDefault() (line 278). In PHP 8+ this throws \Error, not \Exception, so the catch block won't save the day, it'll abort the transaction at flush time.
                                                                                                                                                                                                                           
  Fix both files:                                                                       
                                                                                                                                                                                                                           
  - app/Audit/ConcreteFormatters/SummitScheduleConfigAuditLogFormatter.php:36: getIsDefault() → isDefault()
  - tests/OpenTelemetry/Formatters/SummitScheduleConfigAuditLogFormatterTest.php:58: shouldReceive('getIsDefault') → shouldReceive('isDefault')                                                                            
                                                                                                                                                                                                                           
  The test passes only because Mockery responds to whatever method name you configure — it doesn't validate against the real class.

caseylocker (changes requested):
SummitSchedulePreFilterElementConfigAuditLogFormatter — null config causes uncaught TypeError on pre-filter deletion                                                              
                                                                          
When a pre-filter is removed via SummitScheduleConfig::removePreFilter(), it calls $filter->clearConfig() which nulls the config relation before flush. Since getConfig() has a non-nullable return type (SummitScheduleConfig), PHP 8.3 throws a TypeError when the audit formatter calls it during onFlush. TypeError extends \Error, not \Exception, so neither the formatter's nor the listener's catch block catches it — the entire flush/transaction aborts.

Fix: In SummitSchedulePreFilterElementConfigAuditLogFormatter.php:35-36, guard the null config using the entity's hasConfig() method (available via One2ManyPropertyTrait):

```
$schedule_config = $subject->hasConfig() ? $subject->getConfig() : null;
$config_key = $schedule_config ? ($schedule_config->getKey() ?? 'Unknown Config') : 'Unknown Config';
```

caseylocker (approved):
**APPROVE** — Clean, well-structured PR adding 5 new audit log formatters for SummitDocument, SummitScheduleConfig, SummitSchedulePreFilterElementConfig, SummitSign, and TrackTagGroup.

**Verified**:
- All entity methods called in formatters confirmed to exist on their respective entities
- Follows established AbstractAuditLogFormatter pattern consistently
- Config registrations in `audit_log.php` are correct
- 21 unit tests covering creation/update/deletion/invalid-subject (+ null config edge case for PreFilterElementConfig)
- All 20 CI checks passing
- No security or code quality concerns

smarcet (approved):
LGTM

PR #525, opened 2026-04-09
feat(promo-codes): domain-authorized promo codes for early registration access

caseylocker:
ref: https://app.clickup.com/t/86b952pgc

## Summary

- Implement domain-based early registration access via two new promo code subtypes: `DomainAuthorizedSummitRegistrationDiscountCode` (with discount) and `DomainAuthorizedSummitRegistrationPromoCode` (access-only)
- Add `WithPromoCode` audience value on `SummitTicketType` for promo-code-only ticket types
- Add auto-discovery endpoint (`GET /api/v1/summits/{id}/promo-codes/all/discover`) that finds matching codes for the authenticated user's email
- Add `auto_apply` support to domain-authorized types and existing email-linked types (Member/Speaker)
- Add `QuantityPerAccount` enforcement at both discovery and checkout time
- Comprehensive unit + integration tests covering domain matching, audience filtering, collision avoidance, serialization, discovery, and checkout enforcement

### SDS Implementation (12 tasks)
All 12 tasks implemented. All review follow-ups resolved. Two open deviations remain:
- **D3 (SHOULD-FIX):** `allowed_email_domains` validation needs custom rule (currently `sometimes|json`)
- **D4 (MUST-FIX):** TOCTOU window in `QuantityPerAccount` checkout enforcement — `ApplyPromoCodeTask` needs to move after `ReserveOrderTask` in saga chain

### Files changed (35 files, +3127/-53)
- New models: `DomainAuthorizedSummitRegistrationDiscountCode`, `DomainAuthorizedSummitRegistrationPromoCode`
- New traits: `DomainAuthorizedPromoCodeTrait`, `AutoApplyPromoCodeTrait`
- New interface: `IDomainAuthorizedPromoCode`
- New serializers, factory updates, validation rules, repository queries
- Migration: `Version20260401150000`
- Discovery endpoint: route, controller, service
- Checkout enforcement in `ApplyPromoCodeTask`
- Unit tests: `DomainAuthorizedPromoCodeTest` (30 tests)
- Integration tests: 8 tests in `OAuth2SummitPromoCodesApiTest`

## Test plan

- [ ] `php artisan test --filter=DomainAuthorizedPromoCodeTest` — all unit tests pass
- [ ] `php artisan test --filter="OAuth2SummitPromoCodesApiTest::testDiscover"` — discovery integration tests pass
- [ ] Migration `up` and `down` run cleanly
- [ ] Manual API test: create domain-authorized promo code, hit discover endpoint, verify response
- [ ] Verify existing promo code types are unaffected (regression)

## Summary by CodeRabbit

* **New Features**
  * Domain-authorized promo codes: email-domain eligibility, per-account limits, and remaining-per-account info.
  * Auto-apply flag added to multiple promo/discount code types.
  * Public promo-code discovery endpoint returning codes applicable to the authenticated user.
  * Support for promo-code-only ticket types and inclusion in ticket selection logic.

* **Tests**
  * Extensive unit and API tests covering discovery, domain matching, per-account/global limits, serialization, and validation.


app/Services/Model/Imp/SummitPromoCodeService.php
```diff
@@ -641,6 +642,10 @@ public function importPromoCodes(Summit $summit, UploadedFile $csv_file, ?Member
                     $row['tags'] = explode('|', $row['tags']);
                 }
 
+                if(isset($row['allowed_email_domains'])){
+                    $row['allowed_email_domains'] = explode('|', $row['allowed_email_domains']);
+                }
```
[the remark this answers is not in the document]
caseylocker: ✅ **Addressed in c3f8df754.**

This was caught during our first review pass and fixed in the follow-up commit. Both CSV import paths now trim whitespace, filter empty strings, and unset the key entirely if no valid domains remain — preventing blank cells from bricking promo codes via `matchesEmailDomain()`.

database/migrations/model/Version20260401150000.php
```diff
@@ -0,0 +1,125 @@
...
+
+        // 4. Drop new joined tables
+        $this->addSql("DROP TABLE IF EXISTS DomainAuthorizedSummitRegistrationPromoCode");
+        $this->addSql("DROP TABLE IF EXISTS DomainAuthorizedSummitRegistrationDiscountCode");
+
+        // 4b. Delete orphaned base-table rows before narrowing the ENUM
+        $this->addSql("DELETE FROM SummitRegistrationPromoCode WHERE ClassName IN (
+            'DomainAuthorizedSummitRegistrationDiscountCode',
+            'DomainAuthorizedSummitRegistrationPromoCode'
+        )");
```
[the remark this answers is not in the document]
caseylocker: ✅ **Addressed in c3f8df754.**

This was caught in our first review pass. The `down()` now uses `UPDATE ... SET ClassName = CASE` to remap domain-authorized rows to base types instead of `DELETE`. This preserves FK references — notably `SummitAttendeeTicket.PromoCodeID` which has `ON DELETE CASCADE` and would have silently destroyed ticket history.

database/migrations/model/Version20260401150000.php
```diff
@@ -0,0 +1,131 @@
...
+        $this->addSql("ALTER TABLE SpeakerSummitRegistrationDiscountCode DROP COLUMN AutoApply");
+        $this->addSql("ALTER TABLE SpeakerSummitRegistrationPromoCode DROP COLUMN AutoApply");
+        $this->addSql("ALTER TABLE MemberSummitRegistrationDiscountCode DROP COLUMN AutoApply");
+        $this->addSql("ALTER TABLE MemberSummitRegistrationPromoCode DROP COLUMN AutoApply");
+
+        // 2. Guard against orphaned WithPromoCode values before narrowing the ENUM
+        $this->addSql("UPDATE SummitTicketType SET Audience = 'All' WHERE Audience = 'WithPromoCode'");
+
+        // 3. Revert SummitTicketType Audience ENUM
+        $this->addSql("ALTER TABLE SummitTicketType MODIFY Audience ENUM('All', 'WithInvitation', 'WithoutInvitation') NOT NULL DEFAULT 'All'");
```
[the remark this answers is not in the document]
caseylocker: **Dismissed — intentional design.**

The `down()` is deliberately compatibility-first and lossy. The rollback target (`main`) cannot tolerate the new values:

1. **Doctrine throws on unknown discriminator values.** `main`'s `SummitRegistrationPromoCode` discriminator map only includes the 12 legacy classes. Hydrating a `DomainAuthorized...` row triggers `HydrationException::invalidDiscriminatorValue()`, breaking any code that loads promo codes — including ticket FK hydration.

2. **`main` doesn't understand `WithPromoCode`.** The old `AllowedAudience` array excludes it, and the ENUM narrowing SQL itself cannot succeed while rows contain that value.

3. **Fail-fast is operationally worse.** DDL statements in MySQL commit implicitly — if `down()` aborts partway through, you get a partially migrated schema that neither the old nor new code can run against. The current approach restores a state `main` can actually operate on.

4. **"Inactive state" isn't implementable.** The old code has no concept of inactive domain-authorized types. Any workaround would be more arbitrary than the current remap and still wouldn't solve the discriminator incompatibility.

Independently verified by Codex against the Doctrine ORM hydration path and the FK cascade structure.

app/Models/Foundation/Summit/Factories/SummitPromoCodeFactory.php
```diff
@@ -188,6 +198,8 @@ public static function populate(SummitRegistrationPromoCode $promo_code, Summit
                     $promo_code->setEmail(trim($data['email']));
                 if(isset($data['quantity_available']))
                     $promo_code->setQuantityAvailable(intval($data['quantity_available']));
+                if(isset($data['auto_apply']))
+                    $promo_code->setAutoApply(boolval($data['auto_apply']));
```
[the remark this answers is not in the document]
caseylocker: **Acknowledged — deferred to a separate cleanup.**

Valid finding. `boolval("false")` does return `true` in PHP, and the CSV import path sends raw strings to the factory. However, lines 172/175 show `boolval()` is a **pre-existing pattern** in this factory for `allows_to_delegate` and `allows_to_reassign` — those exist on `main` before this branch. This PR only added the `auto_apply` calls following the same convention.

Fixing only `auto_apply` while leaving the other `boolval()` calls would be inconsistent. A separate cleanup PR should switch all 8 calls to `filter_var($val, FILTER_VALIDATE_BOOLEAN)`.

app/Models/Foundation/Summit/Factories/SummitPromoCodeFactory.php
```diff
@@ -273,6 +291,32 @@ public static function populate(SummitRegistrationPromoCode $promo_code, Summit
                     $promo_code->setQuantityAvailable(intval($data['quantity_available']));
             }
             break;
+            case DomainAuthorizedSummitRegistrationDiscountCode::ClassName:{
+                if(isset($data['allowed_email_domains']))
+                    $promo_code->setAllowedEmailDomains($data['allowed_email_domains']);
```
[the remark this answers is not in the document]
caseylocker: ✅ **Addressed in ae261a7 and c3f8df7.**

Fixed at the correct layer — the CSV import service (`SummitPromoCodeService.php`), not the factory. The service now explodes pipe-delimited `allowed_email_domains`, trims whitespace, filters empty strings, and unsets the key if no valid domains remain. This matches the pattern used for all other CSV list fields (`badge_features`, `allowed_ticket_types`, `tags`, etc.) and keeps the factory clean for API requests which already pass validated arrays.

app/ModelSerializers/Summit/Registration/PromoCodes/MemberSummitRegistrationDiscountCodeSerializer.php
```diff
@@ -83,6 +84,19 @@ public function serialize($expand = null, array $fields = [], array $relations =
...
+            || (!empty($expand) && str_contains($expand, 'allowed_ticket_types'));
+        if ($needs_allowed_ticket_types && !isset($values['allowed_ticket_types'])) {
+            $ticket_types = [];
+            foreach ($code->getAllowedTicketTypes() as $ticket_type) {
+                $ticket_types[] = $ticket_type->getId();
+            }
+            $values['allowed_ticket_types'] = $ticket_types;
+        }
+
+        $values['remaining_quantity_per_account'] = null;
```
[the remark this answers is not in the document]
caseylocker: **Dismissed — correct as implemented.**

`MemberSummitRegistrationDiscountCode` does not have a `quantity_per_account` property or getter. That field exists only on domain-authorized types via `IDomainAuthorizedPromoCode` and `DomainAuthorizedPromoCodeTrait`. The hardcoded `null` correctly indicates "no per-account limit" for member/speaker types. The domain-authorized serializers emit the real value from `getRemainingQuantityPerAccount()`.

Independently verified by Codex — confirmed the member discount model at `MemberSummitRegistrationDiscountCode.php:24` has no per-account state.

app/Rules/AllowedEmailDomainsArray.php
```diff
@@ -0,0 +1,74 @@
...
+            }
+            // .tld — must have at least one char after .
+            elseif (str_starts_with($element, '.')) {
+                if (!preg_match('/^\.\w+$/', $element))
+                    return false;
+            }
+            // user@example.com — standard email-like pattern
+            elseif (str_contains($element, '@')) {
+                if (!preg_match('/^[^@\s]+@[\w][\w.-]+$/', $element))
+                    return false;
```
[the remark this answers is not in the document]
caseylocker: **Dismissed — misframed.**

The domain-matching logic (`matchesEmailDomain()` using `str_ends_with`) handles multi-label suffixes correctly. If `.co.uk` is stored, `user@company.co.uk` matches.

The actual limitation is the **validation regex** in `AllowedEmailDomainsArray.php:51` which only accepts `/^\.\w+$/` — so `.co.uk` can't be submitted via the API. That's a separate enhancement concern, not a bug in the matching logic. PSL integration is not required for the currently implemented behavior.

Independently verified by Codex.

database/migrations/model/Version20260401150000.php
```diff
@@ -0,0 +1,119 @@
...
+            'SpeakerSummitRegistrationPromoCode',
+            'SummitRegistrationDiscountCode',
+            'MemberSummitRegistrationDiscountCode',
+            'SponsorSummitRegistrationDiscountCode',
+            'SpeakerSummitRegistrationDiscountCode',
+            'SpeakersSummitRegistrationPromoCode',
+            'SpeakersRegistrationDiscountCode',
+            'PrePaidSummitRegistrationPromoCode',
+            'PrePaidSummitRegistrationDiscountCode'
+        ) DEFAULT 'SummitRegistrationPromoCode'");
```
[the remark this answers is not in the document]
caseylocker: ✅ **Addressed in ae261a7 (initial fix) and c3f8df7 (improved to UPDATE).**

Originally fixed with a DELETE, then improved to UPDATE after discovering that `SummitAttendeeTicket.PromoCodeID` has `ON DELETE CASCADE` — DELETE would have silently destroyed ticket history. The `down()` now remaps domain-authorized rows to base types via `UPDATE ... SET ClassName = CASE`, preserving all FK references.

routes/api_v1.php
```diff
@@ -1950,6 +1950,10 @@
 
         // promo codes
         Route::group(['prefix' => 'promo-codes'], function () {
+            Route::group(['prefix' => 'all'], function () {
+                // rate-limit only — no authz groups required per SDS Task 9
+                Route::get('discover', ['middleware' => ['rate.limit:25,1'], 'uses' => 'OAuth2SummitPromoCodesApiController@discover']);
+            });
```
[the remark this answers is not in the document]
caseylocker: **Dismissed — authentication is already enforced at two layers.**

1. The route is inside the \`api\` middleware group (\`RouteServiceProvider.php:89\`), which includes \`oauth2.protected\` (\`Kernel.php:59\`). This enforces OAuth2 bearer token authentication.

2. The controller explicitly checks \`getCurrentUser()\` at line 1611 and returns 403 if null.

Adding \`auth.user\` was **intentionally removed** in commit \`138c1f89c\` because \`UserAuthEndpoint\` middleware requires endpoint authz_groups, which caused every request to 403. The route comment explicitly states "no authz groups required per SDS Task 9." The \`rate.limit\` middleware is for abuse prevention, not authentication.

Independently verified by Codex against \`RouteServiceProvider.php\`, \`Kernel.php\`, and \`UserAuthEndpoint.php\`.

app/Models/Foundation/Summit/Factories/SummitPromoCodeFactory.php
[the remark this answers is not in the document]
caseylocker: @coderabbitai yes, open a GitHub issue to track this cleanup so it doesn't get lost.

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitPromoCodesApiController.php
```diff
@@ -1575,4 +1575,66 @@ public function sendSponsorPromoCodes($summit_id)
...
+            if (is_null($summit))
+                return $this->error404();
+
+            $current_member = $this->resource_server_context->getCurrentUser();
+            if (is_null($current_member))
+                return $this->error403();
+
+            $codes = $this->promo_code_service->discoverPromoCodes($summit, $current_member);
+
+            $expand = Request::input('expand', '');
```
smarcet: @caseylocker this not follow api patterns please check other controllers to see how we are doing it 
```PHP
SerializerUtils::getExpand(),
SerializerUtils::getFields(),
SerializerUtils::getRelations()
```

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -278,9 +279,8 @@ private function buildRegularSaga(Member $owner, Summit $summit, array $payload)
         Log::debug(sprintf("SagaFactory::buildRegularSaga - summit id %s", $summit->getId()));
         return Saga::start()
             ->addTask(new PreOrderValidationTask($summit, $payload, $this->ticket_type_repository, $this->tx_service))
-            ->addTask(new PreProcessReservationTask($summit, $payload))
+            ->addTask(new PreProcessReservationTask($summit, $payload, $owner, $this->promo_code_repository))
```
smarcet: @caseylocker this need to be reviewed
what is the rationale of this change ?
seems to me a breaking change without a clear regression test 
Before this PR — saga order was safe                                                                                                                                                              
  ````                                                                                                                                                                                                  
  PreOrderValidationTask                                                                                                                                                                            
    → PreProcessReservationTask (validates ticket types, aggregates promo code usage)                                                                                                            
          → ReserveTicketsTask (decrements ticket type inventory with locks)                                                                                                                  
               → ApplyPromoCodeTask (validates checkSubject, canBeAppliedTo, adds usage)                                                                                                             
                   → ReserveOrderTask (creates Order + Tickets in DB, applies discount via applyTo)  
````
If ApplyPromoCodeTask threw (e.g. checkSubject rejects the user's email domain), the saga compensation ran:

  1. ApplyPromoCodeTask::undo() ( nothing to undo ,no usage was added)
  2. ReserveTicketsTask::undo() (restores ticket type quantities)
  3. PreProcessReservationTask::undo() (no-op)
  4. PreOrderValidationTask::undo() (no-op)

No order existed yet. ReserveOrderTask never ran, so its empty undo() was irrelevant. Clean failure.

**Review by smarcet — CHANGES_REQUESTED** [review]
@caseylocker please review

File: database/seeders/ApiEndpointsSeeder.php [hunk 12]
```diff
@@ -7601,6 +7591,15 @@ private function seedSummitEndpoints()
                     SummitScopes::ReadAllSummitData
                 ]
             ],
+            [
```
smarcet, line 7604: @caseylocker 
need to create a migration for the new endpoint too
check https://github.com/OpenStackweb/summit-api/pull/526/changes#diff-8aedb5d1857bdf6419aebd732cfff1de8ef7352a5bb15cad6831b0ab03cd2014R48

app/Services/Model/Imp/SummitOrderService.php
caseylocker: @smarcet you're right that the rationale wasn't clear and there was no regression test — apologies. Pushed c4bcdef33 which addresses all three of your comments.

**Saga reorder rationale (deviation D4 in the PR description):** the pre-existing order had a TOCTOU gap on `QuantityPerAccount`. `DoctrineSummitRegistrationPromoCodeRepository::getTicketCountByMemberAndPromoCode` filters on `o.Status IN ('Reserved', 'Paid', 'Confirmed')`, so running `ApplyPromoCodeTask` *before* `ReserveOrderTask` meant the count query couldn't see the current checkout's in-flight rows. Two concurrent checkouts could each pass the per-account check and then both add `$qty` tickets. Running `ApplyPromoCodeTask` after the `Reserved` order is persisted — and inside the existing promo-code usage lock — makes the count atomic with respect to concurrent orders.

The reorder exposed two real bugs that I also fixed in this commit:

1. `ReserveOrderTask::undo()` was a TODO stub, so any downstream failure (invalid code / domain reject / `canBeAppliedTo` / `QuantityPerAccount`) was leaving orphaned Order+Ticket rows. Now implemented: detaches tickets from attendees, `Summit::removeOrder`, then `ISummitOrderRepository::delete` (Doctrine `cascade: ['remove']` + `orphanRemoval: true` on `SummitOrder::$tickets` drops the ticket rows).
2. `Event::dispatch(new CreatedSummitRegistrationOrder(...))` was firing inside `ReserveOrderTask::run()`, so listeners would observe orders that later got rolled back. Moved to `SummitOrderService::reserve()` after the full saga succeeds. No listener exists today, but the contract is now correct.

Regression tests in `tests/Unit/Services/SagaCompensationTest.php`: `Saga::abort()` invokes `undo()` in reverse order, and `ReserveOrderTask::undo()` removes the order and detaches tickets.

**Other two comments addressed in the same commit:**
- `discover()` now uses `SerializerUtils::getExpand/getFields/getRelations`.
- Added `database/migrations/config/Version20260412000000.php` following the PR #526 template. FYI it uses the `APIEndpointsMigrationHelper` trait from main — will rebase onto main before your re-review so CI picks it up.

One unrelated bug I noticed while fixing the seeder: the `get-sponsorship` entry had `IGroup::Sponsors` in its `scopes` array instead of `authz_groups`. Fixed in the same commit — happy to split into a separate PR if you'd prefer.

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -662,14 +663,36 @@ public function run(array $formerState): array
                 );
                 $invitation->addOrder($order);
             }
-            Event::dispatch(new CreatedSummitRegistrationOrder($order->getId()));
+            $this->formerState['order'] = $order;
             return ['order' => $order];
```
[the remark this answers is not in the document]
caseylocker: Good catch — verified and fixed in 93bc18027. Before the reorder, `ApplyPromoCodeTask` ran upstream of `ReserveOrderTask`, so `ReserveOrderTask` returning `['order' => $order]` was harmless (nothing downstream consumed the state). After the reorder, `ApplyPromoCodeTask` runs last and reads `formerState['promo_codes_usage']` at line 776 — which would have been undefined. Now returns `$this->formerState` with 'order' merged in.

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitPromoCodesApiController.php
```diff
@@ -1575,4 +1575,63 @@ public function sendSponsorPromoCodes($summit_id)
...
+     * codes (member/speaker, matched by associated email) with auto_apply flag.
+     * Email is always derived from the authenticated principal — no email query parameter accepted.
+     */
+    #[OA\Get(
+        path: "/api/v1/summits/{id}/promo-codes/all/discover",
+        summary: "Discover qualifying promo codes for the current user",
+        description: "Returns domain-authorized promo codes (matched by email domain) and existing email-linked promo codes (member/speaker, matched by associated email) for the current user",
+        operationId: "discoverPromoCodesBySummit",
+        tags: ["Promo Codes"],
+        security: [['summit_promo_codes_oauth2' => [SummitScopes::ReadSummitData]]],
```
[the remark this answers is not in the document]
caseylocker: Fixed in 93bc18027 — added `SummitScopes::ReadAllSummitData` to the security annotation to match the seeded scopes.

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -779,6 +813,26 @@ public function run(array $formerState): array
...
 
+                // QuantityPerAccount enforcement for domain-authorized promo codes
+                // Runs inside the locked transaction, after ReserveOrderTask has created ticket rows
+                if ($promo_code instanceof IDomainAuthorizedPromoCode
+                    && !is_null($this->owner)
+                ) {
+                    $quantityPerAccount = $promo_code->getQuantityPerAccount();
+                    if ($quantityPerAccount > 0) {
+                        $existingCount = $this->promo_code_repository->getTicketCountByMemberAndPromoCode($this->owner, $promo_code);
+                        if ($existingCount > $quantityPerAccount) {
```
romanetar: @caseylocker I think here should be  if ($existingCount >= $quantityPerAccount)

Additionally, the check doesn't account for $qty - eg: a single order for 3 tickets against a             
quantityPerAccount = 2 would bypass the guard entirely.                                                                                                                                     
                                                                                                                                                                                              
We should have sth like this:                                                                                                                                                                                       if (($existingCount + $qty) > $quantityPerAccount)

app/Services/Model/Imp/SummitPromoCodeService.php
```diff
@@ -744,6 +752,13 @@ public function importSponsorPromoCodes(Summit $summit, UploadedFile $csv_file,
                     $row['tags'] = explode('|', $row['tags']);
                 }
 
+                if(isset($row['allowed_email_domains'])){
```
romanetar: @caseylocker there’s some dupe code in lines 645–650 and 755–760. Might be worth extracting it into a private method like parseDelimitedDomains

app/ModelSerializers/Summit/Registration/PromoCodes/DomainAuthorizedSummitRegistrationDiscountCodeSerializer.php
```diff
@@ -0,0 +1,71 @@
...
+     */
+    public function serialize($expand = null, array $fields = [], array $relations = [], array $params = [])
+    {
+        $code = $this->object;
+        if (!$code instanceof DomainAuthorizedSummitRegistrationDiscountCode) return [];
+        $values = parent::serialize($expand, $fields, $relations, $params);
+
+        // RE-ADD allowed_ticket_types (parent discount serializer unsets it).
+        // Check both relations (default serialization) and expand (explicit ?expand= request).
+        $needs_allowed_ticket_types = in_array('allowed_ticket_types', $relations)
```
romanetar: @caseylocker This silently compensates for a parent behavior that's undocumented. If the parent serializer changes, this will break without any error. Consider either overriding the parent behaviour more explicitly (resetting the parent's unset) or documenting why the parent SummitRegistrationDiscountCodeSerializer removes allowed_ticket_types and whether that should be changed there instead.

romanetar (changes requested):
@caseylocker please review comments

app/Services/Model/Imp/SummitOrderService.php
caseylocker: Thanks for catching this — I traced it carefully and I think the current `>` is actually correct, but the code was missing the context to make that obvious, which is fair criticism. Here's the flow:

- `ReserveOrderTask` runs before `ApplyPromoCodeTask` in the saga (`SagaFactory::buildRegularSaga`).
- Inside `ReserveOrderTask::run()` (~line 569) each new ticket has `$promo_code->applyTo($ticket)` invoked, which sets `PromoCodeID`.
- `ReserveOrderTask` wraps its work in `$tx_service->transaction(...)`, which flushes and commits before returning. The new order is persisted in `Reserved` status with `PromoCodeID` set on each ticket.
- `getTicketCountByMemberAndPromoCode` (`DoctrineSummitRegistrationPromoCodeRepository.php:730`) is a raw SQL count over `SummitAttendeeTicket` joined to `SummitOrder` with `o.Status IN ('Reserved','Paid','Confirmed')` and `t.PromoCodeID = :promo_code_id`. So `$existingCount` already includes the current order's tickets.

Under that semantic: `limit=2, buying 2, prior 0 → existingCount=2 → 2 > 2 false → allowed (exactly at cap)`. Switching to `>=` would false-reject this case, and adding `$qty` would double-count.

Pushed 283f57641 which (a) adds an in-depth block comment at the guard explaining the saga-order dependency and why `>` is correct, and (b) adds `tests/Unit/Services/ApplyPromoCodeTaskQuantityPerAccountTest.php` with 6 cases that pin this invariant. Mutation-tested by temporarily flipping to `>=` — the exactly-at-limit case fails as expected. Happy to re-look if you still see a hole.

app/Services/Model/Imp/SummitPromoCodeService.php
caseylocker: Good call, done in 283f57641. Extracted to `private function parseDelimitedDomains(array &$row): void` at the bottom of the class; both call sites (previously 645–650 and 755–760) now just `$this->parseDelimitedDomains($row)`. Behavior is byte-identical.

app/ModelSerializers/Summit/Registration/PromoCodes/DomainAuthorizedSummitRegistrationDiscountCodeSerializer.php
caseylocker: Fair flag. I dug into this — the parent's `unset($values['allowed_ticket_types'])` is actually deliberate: generic discount codes express ticket-type coverage via `ticket_types_rules` (which carries rate/amount per type), so a flat `allowed_ticket_types` would be redundant and ambiguous for a generic discount code. Fully removing the unset at the root would re-introduce that redundancy across all 7 discount-code subclasses — wrong direction.

What was sloppy was that three subclasses (`DomainAuthorized…`, `Member…`, `Speaker…`) each needed the flat list and were independently re-adding it via 8-line duplicates. In 283f57641:

- Added a block comment on the parent's `unset` explaining *why* it's deliberate and pointing subclasses at the helper.
- Added `protected function restoreAllowedTicketTypes(array &$values, $expand, array $relations): void` on the parent.
- Replaced the three duplicated re-add blocks with a single `$this->restoreAllowedTicketTypes(...)` call each.

Wire output is byte-identical; the contract is now explicit so a future edit to the parent can't silently break the children.

caseylocker:
@romanetar thanks for the review — all three comments addressed in 283f57641. Summary:

1. **`SummitOrderService.php:824` (QuantityPerAccount guard)** — the existing `>` is correct because `$existingCount` already includes the in-flight order's tickets (ReserveOrderTask commits before ApplyPromoCodeTask runs; see threaded reply for the full trace). Added an in-depth explanatory comment at the guard + a 6-case unit regression test (`ApplyPromoCodeTaskQuantityPerAccountTest`) that mutation-fails against both `>=` and `+$qty` variants.
2. **`SummitPromoCodeService.php:755` (dupe code)** — extracted `parseDelimitedDomains()` helper and replaced both copies.
3. **`DomainAuthorizedSummitRegistrationDiscountCodeSerializer.php:49` (fragile parent compensation)** — the parent's unset of `allowed_ticket_types` is deliberate (discount codes use `ticket_types_rules`), but three subclasses were all re-adding it with duplicate blocks. Documented the parent's contract and dedup'd via a `protected restoreAllowedTicketTypes()` helper.

No wire behavior changes; `./vendor/bin/phpunit` still green locally.

romanetar (approved):
LGTM

database/migrations/model/Version20260401150000.php
```diff
@@ -0,0 +1,131 @@
...
+ */
+final class Version20260401150000 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema): void
+    {
+        // 1. Create DomainAuthorizedSummitRegistrationDiscountCode joined table
+        $this->addSql("CREATE TABLE DomainAuthorizedSummitRegistrationDiscountCode (
```
smarcet: @caseylocker dont use this there are already utils to create table please review used patterns

database/migrations/model/Version20260401150000.php
```diff
@@ -0,0 +1,131 @@
...
+        $this->addSql("ALTER TABLE SpeakerSummitRegistrationDiscountCode ADD COLUMN AutoApply TINYINT(1) NOT NULL DEFAULT 0");
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema): void
+    {
+        // 1. Drop AutoApply columns from existing email-linked subtype tables
+        $this->addSql("ALTER TABLE SpeakerSummitRegistrationDiscountCode DROP COLUMN AutoApply");
```
smarcet: @caseylocker 
Migration down() Ordering issue
MySQL DDL (DROP TABLE) causes an implicit commit. If the migration fails between step 4 (DROP TABLE) and step 4b (UPDATE ClassName), the joined tables are gone but the
discriminator column still points to them. Doctrine will attempt to join non-existent tables for any row with ClassName DomainAuthorized*, crashing all promo code queries for the entire summit.
Suggested fix:
Swap steps 4 and 4b , remap ClassName to base types first, then drop the joined tables.

app/Services/Model/Imp/SummitOrderService.php
smarcet: @caseylocker TOCTOU bug still exists please review
https://github.com/OpenStackweb/summit-api/pull/530
that PR is a PR from this PR and provide a test that manifest the bug.
race in QuantityPerAccount causes double-rejection of valid concurrent requests
When two requests reserve tickets before either validates, both see an inflated count and both reject ,even though the first was valid. Users lose legitimate purchases under
concurrent load.

The root cause is in SummitOrderService.php:841 — getTicketCountByMemberAndPromoCode counts all committed tickets including the other request's tickets, but the guard has no way to distinguish
  "my order's tickets" from "the concurrent request's tickets."

app/Services/Model/Imp/SummitPromoCodeService.php
```diff
@@ -1008,4 +1013,58 @@ function ($summit, $flow_event, $promocode_id, $test_email_recipient, $announcem
             $filter
         );
     }
+
+    /**
+     * @param Summit $summit
+     * @param Member $member
+     * @return SummitRegistrationPromoCode[]
+     */
+    public function discoverPromoCodes(Summit $summit, Member $member): array
```
smarcet, line 1022: @caseylocker getDiscoverableByEmailForSummit loads all discoverable promo codes into memory
The DQL query fetches every promo code of 6 types for the entire summit, then filters in PHP by domain/email. For summits with thousands of codes, this is a full table scan hydrated into entities.

 If you want to optimize, the cleanest approach is splitting into two queries:

  1. Domain-authorized codes — fetch all for summit, filter in PHP (unavoidable)
  2. Email-linked codes — filter by email in DQL (eliminates the bulk of hydrations)

app/Rules/AllowedEmailDomainsArray.php
```diff
@@ -0,0 +1,74 @@
...
+final class AllowedEmailDomainsArray implements Rule
+{
+    /**
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @return bool
+     */
+    public function passes($attribute, $value): bool
+    {
+        if (!is_array($value))
```
smarcet, line 36: @caseylocker 
AllowedEmailDomainsArray validation rejects multi-level TLD suffix patterns
The .tld suffix regex /^\.\w+$/ rejects .co.uk, .com.au, .ac.uk — common patterns for UK/AU/international organizations. Admins can't create these patterns via the API, even
  though matchesEmailDomain() handles them correctly at runtime.

app/Rules/AllowedEmailDomainsArray.php
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Class AllowedEmailDomainsArray
+ * @package App\Rules
+ *
+ * Validates that the value is an array of allowed email domain patterns.
+ * Each entry must match one of:
+ *   - @domain.com  (exact domain match, starts with @)
+ *   - .tld         (TLD/suffix match, starts with .)
+ *   - user@example.com (exact email match)
+ */
```
smarcet, line 26: @caseylocker 
AllowedEmailDomainsArray validation rule — no tests for edge cases like .co.uk, empty array, mixed valid/invalid patterns

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -662,14 +663,39 @@ public function run(array $formerState): array
...
-            return ['order' => $order];
+            $this->formerState['order'] = $order;
+            // Preserve accumulated state (promo_codes_usage, reservations, etc.) so
+            // ApplyPromoCodeTask — which now runs after ReserveOrderTask — can read
+            // the state populated by PreProcessReservationTask.
+            return $this->formerState;
         });
     }
 
     public function undo()
```
smarcet, line 676: @caseylocker 
ReserveOrderTask::undo() compensation , the new implementation detaches tickets and deletes the order but has no dedicated test verifying the cleanup behavior after ApplyPromoCodeTask failure

smarcet (changes requested):
@caseylocker please review comments

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitPromoCodesApiController.php
caseylocker: Updated in c4bcdef33. The `discover()` method now calls `SerializerUtils::getExpand()/getFields()/getRelations()` (controller L1617-1619), matching the pattern used elsewhere in this controller.

app/Services/Model/Imp/SummitOrderService.php
caseylocker: Addressed. The rationale for the saga reorder now lives alongside the new check location in PreProcessReservationTask::reserveMemberQuotas. The per-account check must run BEFORE ReserveOrderTask commits tickets — post-hoc counting can't distinguish concurrent orders (the TOCTOU race you flagged / PR #530). Regression coverage in SagaCompensationTest and the new PreProcessReservationTaskConcurrencyTest.

database/seeders/ApiEndpointsSeeder.php
caseylocker: Added database/migrations/config/Version20260412000000.php which seeds the discover-promo-codes endpoint + scope associations via APIEndpointsMigrationHelper. Idempotent (WHERE NOT EXISTS), same pattern as PR #526.

database/migrations/model/Version20260401150000.php
caseylocker: Rewrote in b9f258242. Table creation and ADD COLUMN now use the LaravelDoctrine Builder + Table API (matching Version20231208171355 and Version20260407003923). ENUM MODIFY remains raw addSql because Doctrine Schema has no MySQL ENUM support — same pattern as Version20231208172204. Charset aligned with existing promo-code subtype tables (utf8mb3_unicode_ci). Verified with a full down-up-down-up round-trip on the model DB.

database/migrations/model/Version20260401150000.php
caseylocker: Fixed in b9f258242 (same commit as the table-util refactor). The new down() sequence runs the ClassName remap BEFORE the joined-table drops so a mid-rollback failure can't leave the discriminator pointing at dropped tables.

app/Services/Model/Imp/SummitOrderService.php
caseylocker: Addressed by the commit sequence b1f3abea9 through ec9f7301a (9 commits). The root cause you flagged is unfixable with post-hoc counting — deducting the current order's own qty doesn't help either, since two concurrent orders would each subtract their own and still see the other's tickets.

The fix establishes a pre-reservation, durable per-member counter incremented atomically under the existing getByValueExclusiveLock PESSIMISTIC_WRITE lock, inside PreProcessReservationTask (before any tickets are committed):

- New entity SummitPromoCodeMemberReservation(PromoCodeID, MemberID, QtyUsed) with UNIQUE (PromoCodeID, MemberID) and cascading FKs.
- Two migrations — one for the schema (Builder/Table API), one for the backfill from existing committed tickets (GREATEST upsert so re-running is safe).
- PreProcessReservationTask::reserveMemberQuotas opens a short tx per code, acquires the row lock, checks prior + qty > limit, upserts, commits. undo() decrements under the same lock with retry-on-partial-failure (remaining list pattern).
- restoreTicketsPromoCodes (cancel/refund path) now also decrements the reservation counter after successful removeUsage(), closing the counter-leak on legitimate cancellations.
- Post-facto check removed from ApplyPromoCodeTask — that surface can't distinguish concurrent orders.

Your PR #530 test file was cherry-picked as the starting point and then replaced with tests/Unit/Services/PreProcessReservationTaskConcurrencyTest.php (7 cases) plus tests/Unit/Services/RestorePathReservationTest.php (6 cases for the cancel path).

app/Services/Model/Imp/SummitPromoCodeService.php
caseylocker: Split exactly as you suggested in 4db605564 + 999caa1ac. getDiscoverableByEmailForSummit now delegates to two focused methods:

1. getDomainAuthorizedDiscoverableForSummit — fetches only the 2 DA types for the summit, filters by matchesEmailDomain() in PHP (unavoidable pattern match).
2. getEmailLinkedDiscoverableForSummit — 4 DQL queries (one per Member/Speaker x Promo/Discount subtype) push the email filter into the WHERE clause, including the MemberPromoCodeTrait owner-fallback (IS NULL OR empty-string) and the PresentationSpeaker member/registration_request two-hop chain.

Both methods add isLive() date filtering in DQL (:now parameter). The facade method merges results, preserving the existing caller contract — discoverPromoCodes and its exhaustion/quota logic are untouched.

app/Rules/AllowedEmailDomainsArray.php
caseylocker: Widened in c62c0f782. The regex now matches one-or-more dot-prefixed alphanumeric labels (/^\.[a-z0-9]+(?:\.[a-z0-9]+)*$/i), so admins can register .co.uk, .com.au, .ac.uk via the API. Still rejects malformed inputs (., ..com, .com., .co..uk, .-edu). Runtime matcher already accepted these — the validator was incorrectly narrower.

app/Rules/AllowedEmailDomainsArray.php
caseylocker: Added tests/Unit/Rules/AllowedEmailDomainsArrayTest.php in c62c0f782. 31 cases cover the regression fix (.co.uk, .com.au, .ac.uk), valid/invalid pattern mixes, empty array, non-array inputs, whitespace handling, and message(). Uses doc-comment @dataProvider to match the one existing precedent in the repo.

app/Services/Model/Imp/SummitOrderService.php
caseylocker: Covered by tests/Unit/Services/SagaCompensationTest.php::testUndoDeletesOrderAndDetachesTicketsFromAttendees (L100). Asserts the order is deleted and each ticket is detached from its attendee owner after a saga abort.

caseylocker:
## Review Response — All 10 Items Addressed

This push (12 commits, `b9f258242` → `ec9f7301a`) addresses every review item. Here's the map:

### Migration & Pattern Fixes
| # | Item | Commit(s) |
|---|------|-----------|
| 1 | Controller should use `SerializerUtils` | `c4bcdef33` (pre-existing) |
| 3 | New `discover` endpoint needs migration | `Version20260412000000` (pre-existing) |
| 4 | Don't use raw `addSql` in migration | `b9f258242` |
| 5 | Migration `down()` ordering | `b9f258242` |

### TOCTOU Fix (the big one)
| # | Item | Commit(s) |
|---|------|-----------|
| 2 | Saga reorder rationale / regression test | Pre-existing + `PreProcessReservationTaskConcurrencyTest` |
| 6 | TOCTOU race in QuantityPerAccount | `b1f3abea9` → `ec9f7301a` (9 commits) |
| 10 | `ReserveOrderTask::undo` compensation test | `SagaCompensationTest` (pre-existing) |

The TOCTOU fix replaces the unfixable post-facto count with a **durable per-member reservation counter** (`SummitPromoCodeMemberReservation`) incremented atomically under `PESSIMISTIC_WRITE` in `PreProcessReservationTask`, before any tickets are committed. The cancel/refund path (`restoreTicketsPromoCodes`) now decrements this counter alongside `removeUsage()`. PR #530's test was cherry-picked as the starting point. 13 new test cases across two test files.

### Discover Query Optimization
| # | Item | Commit(s) |
|---|------|-----------|
| 7 | `discoverPromoCodes` loads all codes | `4db605564` + `999caa1ac` |

Split into domain-authorized (PHP filter, unavoidable) + 4 targeted DQL queries per email-linked subtype, pushing email matching and `isLive()` date filtering into the database.

### Validation Fix
| # | Item | Commit(s) |
|---|------|-----------|
| 8 | `AllowedEmailDomainsArray` rejects multi-level TLDs | `c62c0f782` |
| 9 | No tests for `AllowedEmailDomainsArray` | `c62c0f782` (31 cases) |

### Test Summary
55 unit tests pass (49 existing + 6 new restore-path tests). The 6 OAuth2 discover integration tests have pre-existing failures (`Unknown column t1.Source`) unrelated to this PR.

app/Services/Model/Imp/SummitOrderService.php
smarcet: @caseylocker there is a simpler fix available than the counter-deduction approach.                                                                                                                
                                                                                                                                                                                                    
  ## Proposed least-invasive fix                                                                                                                                                                    
                                                                                                                                                                                                  
  Hoist the pessimistic lock on the promo code (if provided in payload) to the saga entry point so concurrent requests for the same code serialize **before any ticket is written**.                
                                                                                                                                                                                                    
  Advisory lock at the saga top, keyed on `checkout.summit.{summit_id}.promo.{CODE}.lock` — keyed by code string (not ID) because the entity isn't resolved until `PreProcessReservationTask`. Same 
  mechanism as the existing `promocode.{id}.usage.lock` at line 856, hoisted to cover the full checkout window. No transaction boundary changes.
                                               
  ## Action

  If you agree, rebase out the commits related to the latest proposed TOCTOU fix (counter-deduction approach) and implement this instead:

  ```bash
  git rebase -i <base>   # drop the TOCTOU-fix commits
  git push --force-with-lease   # NOT --force, so nobody else's pushes get clobbered
  ```

  ## Implementation

  ### 1. Add three helpers to `SummitOrderService`

  ```php
  /**
   * Extract the distinct, normalized promo codes referenced by a reservation
   * payload. Codes are uppercased + trimmed to match the normalization applied
   * by PreProcessReservationTask so the same lock covers the same code no
   * matter the casing in the client payload.
   *
   * @param array $payload
   * @return array<string>
   */
  private function extractPromoCodes(array $payload): array
  {
      $codes = [];
      $tickets = $payload['tickets'] ?? [];
      foreach ($tickets as $ticket_dto) {
          $code = $ticket_dto['promo_code'] ?? null;
          if (empty($code)) continue;
          $codes[strtoupper(trim($code))] = true;
      }
      return array_keys($codes);
  }

  /**
   * Run $callback while holding a pessimistic advisory lock for each distinct
   * promo code in $codes. Locks are acquired in sorted order to avoid
   * lock-ordering deadlocks when a reservation uses more than one code.
   *
   * @param int $summit_id
   * @param array<string> $codes
   * @param \Closure $callback
   * @return mixed
   */
  private function runWithPromoCodeLocks(int $summit_id, array $codes, \Closure $callback)
  {
      if (empty($codes)) {
          return $callback();
      }
      sort($codes);
      return $this->acquirePromoCodeLocksAndRun($summit_id, $codes, 0, $callback);
  }

  /**
   * @param int $summit_id
   * @param array<string> $codes
   * @param int $index
   * @param \Closure $callback
   * @return mixed
   */
  private function acquirePromoCodeLocksAndRun(int $summit_id, array $codes, int $index, \Closure $callback)
  {
      if ($index >= count($codes)) {
          return $callback();
      }
      $lock_name = sprintf('checkout.summit.%s.promo.%s.lock', $summit_id, $codes[$index]);
      return $this->lock_service->lock(
          $lock_name,
          fn() => $this->acquirePromoCodeLocksAndRun($summit_id, $codes, $index + 1, $callback)
      );
  }
  ```

  ### 2. Wrap the saga run in `reserve()`

  ```php
  public function reserve(?Member $owner, Summit $summit, array $payload): SummitOrder
  {
      // ...

      // Saga-level pessimistic lock on each promo code used in this reservation.
      // Serializes concurrent checkouts that share a promo code so per-account
      // enforcement in ApplyPromoCodeTask reads a stable, post-commit ticket count
      // without TOCTOU interleaving from another in-flight order. Locks are
      // acquired in deterministic (sorted) order across codes to avoid lock-
      // ordering deadlocks when an order uses more than one code.
      $promo_codes = $this->extractPromoCodes($payload);
      $state = $this->runWithPromoCodeLocks(
          $summit->getId(),
          $promo_codes,
          fn() => $saga_factory->build($owner, $summit, $payload)->run()
      );

      // Dispatch only after the full saga (including ApplyPromoCodeTask validation) succeeds,
      // so listeners never observe orders that were rolled back by compensation.
      Event::dispatch(new CreatedSummitRegistrationOrder($state['order']->getId()));

      return $state['order'];
  }
  ```

  ### 3. Replace the fragile comment in `ApplyPromoCodeTask::run()`

  The current 20-line comment explaining the `>` vs `>=` landmine and the saga-ordering invariant is no longer needed and is actively misleading — the semantics are now protected by the saga-level
   lock, not by the fragile timing of `ReserveOrderTask` committing first. Replace with:

  ```php
  // QuantityPerAccount enforcement for domain-authorized promo codes.
  //
  // Concurrent checkouts for the same promo code are serialized by the
  // saga-level advisory lock acquired in SummitOrderService::reserve,
  // so $existingCount reflects the true committed state for this
  // member at the point of check. The condition is `>` (not `>=`)
  // and we do NOT add $qty because ReserveOrderTask already committed
  // this order's tickets with PromoCodeID set via applyTo(), and
  // getTicketCountByMemberAndPromoCode includes them.
  if ($promo_code instanceof IDomainAuthorizedPromoCode
      && !is_null($this->owner)
  ) {
      // ... existing guard body unchanged
  }
  ```

  ### 4. Add regression test

  Add `tests/Unit/Services/SummitOrderServiceReservePromoCodeLockTest.php` covering:

  - `extractPromoCodes()` normalizes (uppercase + trim) and deduplicates
  - `extractPromoCodes()` returns `[]` when no tickets carry a `promo_code`
  - `runWithPromoCodeLocks()` skips `$lock_service->lock` entirely when no codes are present (no regression for non-promo orders)
  - Single-code case acquires exactly one lock with the expected name
  - Multi-code case acquires all locks in sorted order, releases in inverse order (deadlock-safe)
  - Exceptions thrown inside the callback propagate through the locks

  The helpers can be exercised via reflection so the test doesn't need the full `SummitOrderService` dependency graph (Doctrine, repositories, etc.).

  ## PR #530

  After this fix lands, the unit test `testDoubleRejection_BothReservedBeforeEitherValidates` in PR #530 still fails when run in isolation — it exercises `ApplyPromoCodeTask` alone, and the
  saga-level lock lives one layer up in `SummitOrderService::reserve`. Two options, pick one:

  1. **Remove it** — the scenario cannot occur in the real flow anymore.
  2. **Keep as negative documentation** — rename to make clear it documents what would break without the saga lock; skip it with `$this->markTestSkipped('documents historical TOCTOU; serialized by
   saga lock in SummitOrderService::reserve')`.

  My vote is (1); the bug is closed at the integration boundary, and keeping a red test around for documentation has a way of rotting.

  ## Why this is the minimal fix

  - Reuses existing `ILockManagerService` (same pattern as `promocode.{id}.usage.lock` at line 856)
  - No counter-deduction logic, no injected state, no saga refactor, no new interfaces
  - No changes to `ApplyPromoCodeTask` logic — only the misleading comment is replaced
  - Deterministic lock ordering handles the multi-code-per-order case safely
  - Failure mode under contention is the same as the existing usage lock (`UnacquiredLockException` after 3 retries) — no new operational surprise
  - With the lock in place, the `testDoubleRejection_BothReservedBeforeEitherValidates` scenario from PR #530 cannot occur at the saga entry point because concurrent requests for the same code
  serialize before `ReserveOrderTask` commits any ticket

caseylocker: @smarcet thanks for the detailed write-up — the instinct to reuse `ILockManagerService` is reasonable. After working through it (and independent review from Codex), I'd like to keep the counter-based approach on this branch. Here's the reasoning.

**1. `LockManagerService` is not safe for saga-wide serialization as currently implemented.**

This is the load-bearing objection. `LockManagerService::lock()` at `app/Services/Utils/LockManagerService.php:86` always calls `releaseLock()` in its `finally` block — including when `acquireLock()` throws `UnacquiredLockException`. Acquisition is a Redis `SETNX` (`RedisCacheService.php:306`); release is an unconditional `DEL` (`RedisCacheService.php:175`), with no ownership token / compare-and-delete check on release.

Scenario under contention:
1. Alice acquires the saga lock → enters her checkout.
2. Bob retries 3× (~500 ms envelope), hits `UnacquiredLockException`.
3. Bob's `finally` runs `releaseLock()` → `DEL` → **Alice's key is gone**.
4. Carol arrives, `SETNX` succeeds, enters her checkout concurrently with Alice.

Mutual exclusion is lost exactly in the regime where it matters. The existing `promocode.{id}.usage.lock` at line 828 has the same bug, but is held for milliseconds, so contention windows are materially narrower; using the same primitive to bracket a full saga (seconds) makes the bug reachable under normal load.

Before saga-level advisory locking is viable at all — here or as a follow-up — `LockManagerService` needs a fix: at minimum, only release when acquisition actually succeeded (gate `releaseLock()` on an `$acquired` flag); ideally, move to token-based compare-and-delete semantics so a release only affects a key the caller still owns.

**2. Critical-section scope.**

Independent of the bug above, the proposal extends a millisecond-scoped lock pattern (`addUsage` at `SummitOrderService.php:828`) to wrap the full `$saga_factory->build(...)->run()`. The current counter-based fix keeps the critical section narrow: row-lock the promo code, check/increment `SummitPromoCodeMemberReservation`, commit, then continue. MySQL row locks aren't infallible — they can deadlock or time out — but at this scope they're the better primitive: they typically wait up to the engine timeout.

**3. Honest trade-off on durable state.**

`SummitPromoCodeMemberReservation` (defined in `Version20260415191521`, backfilled from history in `Version20260415191522`) is a denormalized per-member counter, not a full audit trail. It gives us O(1) reads inside the existing `getByValueExclusiveLock` transaction, and the same state is surfaced during discovery (`SummitPromoCodeService.php:1038`,
`DomainAuthorizedSummitRegistrationPromoCodeSerializer.php:43`). A derived-count approach can still surface remaining allowance during discovery, but it has to compute it from ticket queries rather than reading persisted quota state. The cost of the counter is maintenance: [decrement on saga undo](app/Services/Model/Imp/SummitOrderService.php#L1236), [decrement on order cancel](app/Services/Model/Imp/SummitOrderService.php#L2792), and the risk of drift from ticket state under an unknown failure mode. A derived count can't drift, but scales worse on hot codes.

**Where your proposal is clearly right:**
- Cancel-path decrement is complexity the entity introduces, not a gap the advisory lock leaves.
- Counter drift is a real risk class the derived-count approach doesn't have.
- Operationally lighter on the data layer: one less table, one less migration pair.

**Path forward:**
I'd like to keep the counter work on this branch. If you want saga-level advisory locking as defense-in-depth down the line, that's a worthwhile conversation — but it needs `LockManagerService` fixed first. Happy to open a separate PR for the lock-safety fix if you agree it's a problem, independent of this PR's merge.

Focused tests for the current fix (`PreProcessReservationTaskConcurrencyTest`, `SagaCompensationTest`, `RestorePathReservationTest`) are 16/16 green.

**THAT** all said, let me know and I'll implement your suggestion verbatim if you still think it's better. Just wanted to get your take on my evaluation first.

smarcet: @caseylocker 
honestly i do think that there is still a simpler approach, ( i do agree with lock manager issue by the way):
instead of introducing an external locking mechanism + reservation tracking, we could rely on DB-level guarantees and keep everything inside a single transactional boundary:
```PHP
// SummitOrderService::reserve
$this->tx_service->transaction(function () use ($owner, $summit, $payload) {
// Acquire row-level locks for all promo codes involved (sorted to avoid deadlocks)
foreach ($sorted_codes as $code) {
    $this->promo_code_repository->getByValueExclusiveLock($summit, $code);
}
```
// Run the full saga inside the same transaction
return $saga_factory->build($owner, $summit, $payload)->run();});
This gives us:

Real mutual exclusion via InnoDB row locks (owned by the transaction, no external state like Redis tokens)
A correct per-account constraint using a simple COUNT(*) within the same tx (no drift between check and write)
Automatic rollback semantics — if anything fails, everything is reverted without needing extra compensation logic
Consistent read model (discovery queries match the same source of truth by construction)
Conceptually, this removes the need for an additional reservation layer and keeps the invariant enforcement closer to the data.
Before going in that direction, I think the only things we’d need to validate are:

No external I/O inside the saga transaction (from a quick pass, reserve-path tasks look DB-only)
Nested transaction behavior (ensuring savepoints are enabled)
Deterministic lock ordering (to avoid deadlocks with multiple promo codes)
Lock wait handling (surfacing timeouts properly)


I’m not opposed to your approach , just want to make sure we’re not adding extra moving parts if the DB can already give us the guarantees we need.

smarcet: @caseylocker 
Ok, after reviewing this more carefully:
The transaction manager would need to properly support nested transactions, which is out of scope for this PR.
We could remove the inner transactions in each saga task and rely on a single outer transaction, but that comes with a trade-off I’m not willing to take , increased lock contention, especially under high concurrency (e.g., flash sale scenarios), which could impact system stability.
Given that, I’m OK moving forward with your approach for now, since it results in shorter lock scopes and lower contention.
We can revisit the transactional approach once the transaction handling layer is improved.

smarcet (approved):
LGTM

PR #529, opened 2026-04-15
fix(sponsor-sync): use SELECT result to detect row existence in addSponsorPermission, not UPDATE affected-row count

romanetar:
ref https://app.clickup.com/t/86b94hpua

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sponsor permission tracking to correctly handle cases where permissions already exist, ensuring consistent behavior when permissions are re-applied or after retry operations.


smarcet (approved):
LGTM

PR #530, opened 2026-04-15
chore(unit-test): add unit test to demo the toctou bug

smarcet:
The output is .E. — tests 1 and 3 pass, test 2 (testDoubleRejection_BothReservedBeforeEitherValidates) fails with: ValidationException: Promo code DOMAIN-CODE-1 has reached the maximum of 1 tickets per account. Task A throws at SummitOrderService.php:843 (the $existingCount > $quantityPerAccount guard) (exactly the TOCTOU bug). The test asserts Task A should succeed (it's a valid first request), but the inflated count from both orders' tickets being visible causes it to reject. When the race condition is fixed, this test will start passing.

### DIFF
--- a/tests/Unit/Services/ApplyPromoCodeTaskConcurrencyTest.php
+++ b/tests/Unit/Services/ApplyPromoCodeTaskConcurrencyTest.php
@@ -1,0 +1,328 @@
+<?php namespace Tests\Unit\Services;
+/**
+ * Copyright 2026 OpenStack Foundation
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ **/
+
+use App\Services\Model\ApplyPromoCodeTask;
+use App\Services\Utils\ILockManagerService;
+use libs\utils\ITransactionService;
+use models\exceptions\ValidationException;
+use models\main\Member;
+use models\summit\IDomainAuthorizedPromoCode;
+use models\summit\ISummitRegistrationPromoCodeRepository;
+use models\summit\ISummitRepository;
+use models\summit\Summit;
+use models\summit\SummitRegistrationPromoCode;
+use models\summit\SummitTicketType;
+use Mockery;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Simulates concurrent QuantityPerAccount enforcement in ApplyPromoCodeTask
+ * by controlling what getTicketCountByMemberAndPromoCode returns for each call.
+ *
+ * The race condition: two ReserveOrderTask executions commit tickets before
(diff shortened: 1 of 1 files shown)


smarcet:
Test 1 — testSerializedExecution_FirstSucceeds_SecondRejects (passes)
  - Mock sequencing via andReturnValues([1, 2]) correctly simulates serialized DB reads.
  - addUsage->once() is strict and correct — only Task A reaches it.
  - lock mock has no count constraint (Mockery defaults to zeroOrMoreTimes), which is correct since only Task A reaches it.
  - expectException placed immediately before Task B's run() — no assertion gap.
  - Severity: none

  Test 2 — testDoubleRejection_BothReservedBeforeEitherValidates (fails — by design)
  - Asserts correct behavior (Task A succeeds), which fails against the buggy code — demonstrating the TOCTOU race.
  - try/catch + $this->fail() produces a clean PHPUnit Failure (F) with a descriptive message, not a raw Error (E) stack trace.
  - addUsage->zeroOrMoreTimes() and lock->zeroOrMoreTimes() are permissive — no secondary Mockery errors in tearDown regardless of which code path executes.
  - andReturn(2) (not andReturnValues) correctly returns the same inflated count for every call — both tasks see identical state.
  - When the TOCTOU fix lands, the try/catch path is skipped, assertTrue on line 226 verifies Task A succeeded, and expectException on line 233 verifies Task B rejects. The test starts passing
  with zero changes needed.
  - Severity: none

  Test 3 — testSerializedExecution_BothAllowedWithinLimit (passes)
  - andReturnValues([1, 2]) with limit=2: both 1 > 2 and 2 > 2 are false — both pass.
  - addUsage->twice() is strict and correct — both tasks reach it.
  - Mirrors Test 1 structure but confirms the boundary at count == limit (allowed by > not >=).
  - Severity: none

  Shared infrastructure (setUp/tearDown/bindSummitRepository)
  - Identical to ApplyPromoCodeTaskQuantityPerAccountTest — consistent pattern across the test suite.
  - Severity: none

  Summary

  ┌─────┬────────────────────────────────────────────┬────────┬─────────────────────────────────────────────────────┐
  │  #  │                    Test                    │ Status │                       Purpose                       │
  ├─────┼────────────────────────────────────────────┼────────┼─────────────────────────────────────────────────────┤
  │ 1   │ Serialized: first succeeds, second rejects │ Pass   │ Correct behavior under FOR UPDATE serialization     │
  ├─────┼────────────────────────────────────────────┼────────┼─────────────────────────────────────────────────────┤
  │ 2   │ Double rejection (TOCTOU)                  │ Fail   │ Demonstrates the race — will pass when bug is fixed │
  ├─────┼────────────────────────────────────────────┼────────┼─────────────────────────────────────────────────────┤
  │ 3   │ Serialized: both within limit              │ Pass   │ Boundary case at count == limit with two requests   │
  └─────┴────────────────────────────────────────────┴────────┴─────────────────────────────────────────────────────┘

PR #536, opened 2026-04-29
fix(db): convert SponsorBadgeScan.Notes to utf8mb4

smarcet:
ref: https://app.clickup.com/t/86b9nvmdj
The Notes column is on utf8 (3-byte), causing 'Incorrect string value' PDOExceptions when users submit 4-byte characters (emoji, supplementary plane). Convert to utf8mb4_general_ci to match the precedent set for SummitAttendee.AdminNotes in Version20221012133111.


romanetar (approved):
LGTM

PR #537, opened 2026-05-04
fix(lock): implement Redlock single-instance pattern in LockManagerService

romanetar:
ref https://app.clickup.com/t/86b9f3a22

Recommended actions for a follow-up ticket:                                                                                                                              
  1. Move the SendAttendeeInvitationEmail::dispatch call outside the lock callback (dispatch after the lock is released).
  2. Consider a tighter explicit lifetime (e.g. 30 s) that matches the realistic worst-case DB write time rather than the 3600 s default.                                                     
  3. Fix the missing . in the key: 'ticket_type.' . $type_id . '.promo_code.' . $promo_code_val . '.sell.lock'.  

## Summary by CodeRabbit

* **New Features**
  * Added atomic compare-and-delete support for cache entries.
  * Locks now use ownership tokens, retry with backoff, and prevent unauthorized releases.
* **Bug Fixes**
  * Improved cache initialization and expiration handling.
  * Updated promo-code and reservation locking with explicit 30-second timeouts and corrected lock coordination.
* **Tests**
  * Added coverage for lock ownership, Redis outages, cache expiration, single-value additions, and token-mismatch protection.


tests/Unit/Services/LockManagerServiceOwnershipTest.php
```diff
@@ -0,0 +1,156 @@
...
+        $service->lock('test.lock', function () {}, 3600);
+
+        // Tokens cleared — confirms the single addSingleValue call was paired
+        // with exactly one deleteIfValueMatches (not a separate expire call).
+        $ref  = new ReflectionClass($service);
+        $prop = $ref->getProperty('tokens');
+        $prop->setAccessible(true);
+        $this->assertEmpty($prop->getValue($service));
+    }
+}
```
smarcet, line 181: **Missing test: Redis unavailable during `releaseLock`**

There is no test for the case where `deleteIfValueMatches` returns `false` (Redis connection fails). When that happens the lock is **not deleted from Redis** but `unset($this->tokens[$name])` still runs — the local token handle is gone. The lock persists in Redis until TTL expiry with no way to retry the release from application code.

Suggested outline:
```php
public function testReleaseLockWhenRedisDownLeavesKeyUntilTtl(): void
{
    $cache = Mockery::mock(ICacheService::class);
    $cache->shouldReceive('addSingleValue')->once()->andReturn(true);
    // simulate Redis unavailable — deleteIfValueMatches returns false
    $cache->shouldReceive('deleteIfValueMatches')->once()->andReturn(false);

    $lock = new LockManagerService($cache);
    $lock->lock('resource.lock', fn() => null);
    // Observable effect: a subsequent acquireLock on the same key will fail
    // because the key was NOT deleted — only TTL expiry can free it.
}
```
Documenting this as a test makes the failure mode explicit for future readers.

app/Services/Utils/RedisCacheService.php
```diff
@@ -306,12 +306,11 @@ public function setSingleValue($key, $value, $ttl = 0)
...
-            if ($res && $ttl > 0) {
-                $conn->expire($key, $ttl);
+            if ($ttl > 0) {
+                return $conn->set($key, $value, 'EX', (int)$ttl, 'NX') !== null;
             }
-            return $res;
-        });
+            return $conn->set($key, $value, 'NX') !== null;
+        }, false);
     }
```
smarcet, line 316: **Missing integration test: atomic `SET NX EX`**

All tests for this path mock `ICacheService`, so the actual Redis command added here is never exercised against a real server. Two things that only an integration test can verify:

1. **Driver compatibility** — the variadic form `set($key, $value, 'EX', $ttl, 'NX')` works with Predis 2.x. If the driver is ever switched to PhpRedis, `set()` returns `false` on an NX-miss (not `null`), silently breaking the `!== null` check.
2. **Atomicity** — the fix guarantees the key cannot exist without a TTL; only a real `TTL key` call after `addSingleValue` can confirm there is no gap.

Recommend a `@group integration` test that:
- calls `addSingleValue($key, $token, 30)` against a real test Redis
- reads `TTL $key` and asserts it is between 1 and 30 s
- calls `addSingleValue` again on the same key and asserts it returns `false` (NX semantics)

app/Services/Utils/LockManagerService.php
```diff
@@ -46,22 +50,24 @@ public function __construct(ICacheService $cache_service){
      */
     public function acquireLock(string $name, int $lifetime = 3600):LockManagerService
```
smarcet: **Default TTL of 3600 s creates hour-long stuck locks in queue workers**

This service is resolved inside long-lived queue workers (`ProcessPaymentGatewayRefundJob`, `ProcessSummitOrderPaymentConfirmation`, `RevokeSummitOrder`, etc.). If the worker is killed (`SIGKILL`, OOM, deployment restart) while holding a lock, the `finally` block never executes and the Redis key persists for 3600 s — blocking ticket sales on the affected key for up to **one hour**.

The PR description already flags this: *"Consider a tighter explicit lifetime (e.g. 30 s) that matches the realistic worst-case DB write time."*

Suggested fix — pass an explicit `$lifetime` at the call sites rather than relying on the default:
```php
// SummitOrderService.php:951, 968, 1539
$this->lock_service->lock(
    'ticket_type.' . $ticket_type->getId() . '.sell.lock',
    function () use (...) { ... },
    30  // worst-case DB write time in seconds
);
```

app/Services/Utils/LockManagerService.php
```diff
@@ -22,14 +22,18 @@
...
-    const BackOffBaseInterval = 100000; // 1 ms
+    const BackOffBaseInterval = 100000; // microseconds
+
     /**
      * @var ICacheService
      */
     private $cache_service;
 
+    /** @var array<string,string>  lock-name → per-call ownership token */
+    private array $tokens = [];
```
smarcet: **Mutable per-acquisition state on a singleton service**

`LockManagerService` is bound as a singleton (`BaseServicesProvider.php:143`). Before this PR the service was stateless; this line introduces the first mutable instance state. In the current stack (PHP-FPM, synchronous queue workers) this is safe because only one coroutine runs at a time per process. However, if Laravel Octane / Swoole is ever introduced, two concurrent coroutines racing on the same key name would overwrite each other's token here — coroutine A's `releaseLock` would present coroutine B's token to `deleteIfValueMatches`, releasing B's live lock.

Suggested mitigation: either change the binding to `scoped` (Octane-safe per-request singleton), or remove instance state entirely by returning the token from `acquireLock` and accepting it as a parameter in `releaseLock`.

app/Services/Utils/RedisCacheService.php
```diff
@@ -239,7 +239,7 @@ public function storeHash($name, array $values, $ttl = 0)
     public function incCounter($counter_name, $ttl = 0)
     {
         return $this->retryOnConnectionError(function ($conn) use ($counter_name, $ttl) {
-            if ($conn->setnx($counter_name, 1)) {
+            if ($conn->set($counter_name, 1, ['NX' => true]) !== null) {
```
smarcet: **`incCounter` partial fix — TTL assignment still non-atomic**

`addSingleValue` was made fully atomic (`SET key value EX $ttl NX` in one command), but `incCounter` only swaps `setnx` for `SET NX` while keeping the `expire` as a separate call. If the process crashes between these two commands the counter key will exist indefinitely with no TTL — the same race that `addSingleValue` was fixed to prevent.

Suggested fix — collapse into a single atomic command, consistent with the approach used in `addSingleValue`:
```php
if ($conn->set($counter_name, 1, ['EX' => (int)$ttl, 'NX' => true]) !== null) {
    return 1;
}
```

smarcet:
**Malformed lock key — missing dot separator** ()

This line cannot be commented inline since `SummitOrderService.php` is not in the diff, but the issue is worth tracking here.

The lock key is missing a `.` between `$type_id` and `'promo_code.'`:

```php
// Current — produces e.g. "ticket_type.42promo_code.SUMMER25.sell.lock"
$this->lock_service->lock('ticket_type.' . $type_id . 'promo_code.' . $promo_code_val . '.sell.lock', ...)

// Intended — "ticket_type.42.promo_code.SUMMER25.sell.lock"
$this->lock_service->lock('ticket_type.' . $type_id . '.promo_code.' . $promo_code_val . '.sell.lock', ...)
```

With integer type IDs there is no key collision today, but the key is semantically malformed and will confuse any tooling (monitoring, manual Redis inspection, key-expiry scripts) that parses the key pattern. There is also no test asserting the exact key string passed to `ILockManagerService::lock()` in this code path — a mock expectation on the key argument would catch this immediately.

The PR description already flags this as a follow-up: *"Fix the missing . in the key."* Recommend addressing it in a dedicated follow-up ticket before the ownership token changes are deployed, so the key format stabilises.

smarcet (changes requested):

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -1658,7 +1658,7 @@ function () use ($promo_code_val, $type_id) {
 
 
                     return $order;
-                });
+                }, 30);
```
smarcet, line 1677: @romanetar This 30 s lifetime makes TTL expiry a silent loss of mutual exclusion, and this call site is the only one of the four where the Redis lock is the *sole* guard of the invariant — the fix should be a DB pessimistic lock, matching the sibling tasks.

The other three sites are already backed by row locks — `ReserveTicketsTask::run/undo` hold `getByIdsExclusiveLock`/`getByIdExclusiveLock` on the ticket types and `RedeemPromoCodesTask` holds `getByValueExclusiveLock` on the promo code — so the DB serializes competing writers regardless of what Redis does. This callback instead reads the promo code via `$this->summit->getPromoCodeByCode()` with no lock, and the check-then-act sequence (`hasPrePaidTicketsAssignedBy` → `getNextAvailableTicketPerType` → `addTicket`) relies entirely on this Redis lock. If the callback ever outlives the 30 s TTL, a concurrent request acquires the lock and can hand the same prepaid ticket to two attendees. This failure mode is new to this PR: on `main` the key effectively never expired (TTL was a unix timestamp, ≈55 years). There is also a second window the Redis lock cannot cover even when it works: it is released when the callback returns, but the enclosing `tx_service->transaction` commits afterwards, so a competitor can read pre-commit state.

Suggested fix — same pattern as `RedeemPromoCodesTask` in this file: inject `ISummitRegistrationPromoCodeRepository` into `AutoAssignPrePaidTicketTask` and fetch the promo code with `getByValueExclusiveLock($this->summit, $promo_code_val)` inside the transaction, instead of the unlocked `getPromoCodeByCode()`. That method takes `PESSIMISTIC_WRITE` with `HINT_REFRESH` (DoctrineSummitRegistrationPromoCodeRepository.php:617), so the entity is re-hydrated from committed DB truth after the row lock is granted, and the lock is held through commit — closing both the TTL-expiry hole and the release-before-commit window in one move. The Redis lock can stay as a fail-fast layer, consistent with the sibling sites.

Independently, `releaseLock()` discarding the `deleteIfValueMatches() === false` signal still deserves the log + metric suggested in the CodeRabbit thread on `LockManagerService.php:79` — a mismatch at release is the only runtime evidence that a lock expired mid-callback anywhere in the system.

app/Services/Utils/RedisCacheService.php
```diff
@@ -239,8 +239,7 @@ public function storeHash($name, array $values, $ttl = 0)
     public function incCounter($counter_name, $ttl = 0)
     {
         return $this->retryOnConnectionError(function ($conn) use ($counter_name, $ttl) {
-            if ($conn->setnx($counter_name, 1)) {
-                if ($ttl > 0) $conn->expire($counter_name, (int)$ttl);
+            if ($conn->set($counter_name, 1, ['EX' => (int)$ttl, 'NX' => true]) !== null) {
```
smarcet: @romanetar `incCounter` now fails on every invocation under the Redis client this app actually uses.

The options-array form `set($key, 1, ['EX' => $ttl, 'NX' => true])` is phpredis API, but the app pins Predis (`config/database.php:176`, `REDIS_CLIENT` default `predis`), and Predis's `SET` command passes arguments through verbatim — `StreamConnection::writeRequest` serializes each argument with `strlen(strval($argument))`. With an array argument that's an "Array to string conversion": under Laravel's error handler it throws `ErrorException`; otherwise the wire command becomes `SET key 1 "Array"` and Redis replies `ERR syntax error` (`Predis\Response\ServerException`). Neither exception is caught by `retryOnConnectionError`, which only catches `PredisConnectionException|\RedisException`. Additionally, the default `$ttl = 0` would produce `EX 0`, which Redis rejects even in the correct form.

`addSingleValue` a few lines below uses the correct Predis variadic form — mirroring it fixes both problems:

```php
if ($ttl > 0) {
    if ($conn->set($counter_name, 1, 'EX', (int)$ttl, 'NX') !== null) return 1;
} else {
    if ($conn->set($counter_name, 1, 'NX') !== null) return 1;
}
return (int)$conn->incr($counter_name);
```

`incCounter` has no callers in the repo today, so reverting this hunk entirely is also reasonable — it's outside the lock scope of this PR.

tests/Integration/RedisCacheServiceAddSingleValueTest.php
```diff
@@ -0,0 +1,99 @@
...
+ *     PhpRedis, set() returns false on an NX-miss (not null), which would
+ *     silently break the `!== null` check; this test catches that regression.
+ *
+ *  2. Atomicity — key and TTL are written in a single command; there is no
+ *     window where the key exists without a TTL.  Verified by reading TTL
+ *     immediately after addSingleValue returns.
+ *
+ */
+#[Group("integration")]
+final class RedisCacheServiceAddSingleValueTest extends TestCase
```
smarcet, line 44: @romanetar This test never runs in CI: the `push.yml` suite matrix (`.github/workflows/push.yml:57-68`) is a fixed list of paths/filters with no entry for `tests/Integration/`, so the driver-compatibility and atomicity regressions this file exists to catch would pass silently. Redis is already provisioned in that workflow (`supercharge/redis-github-action`, line 148), so wiring it up is one line:

```yaml
- { name: "Integration", filter: "tests/Integration/" }
```

While extending this class, please also add a case for `deleteIfValueMatches` — it's the Lua compare-and-delete the whole ownership guarantee rests on, and today it's only ever exercised against mocks. A broken script (or a driver change breaking the Predis-style `eval($lua, 1, $key, $value)` signature) would silently no-op every release, holding all locks to full TTL. Two assertions suffice: matching token deletes the key; non-matching token leaves it intact.

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -1536,7 +1536,7 @@ public function run(array $formerState): array
             if (empty($promo_code_val)) throw new ValidationException("Promo code is required.");
 
             $type_id = $ticket_dto['type_id'];
-            $order = $this->lock_service->lock('ticket_type.' . $type_id . 'promo_code.' . $promo_code_val . '.sell.lock',
+            $order = $this->lock_service->lock('ticket_type.' . $type_id . '.promo_code.' . $promo_code_val . '.sell.lock',
```
smarcet, line 1555: @romanetar The key fix is correct, but note it splits the lock namespace between fleets during a rolling deploy: old pods lock `ticket_type.7promo_code.X.sell.lock` while new pods lock `ticket_type.7.promo_code.X.sell.lock`, so the two fleets don't mutually exclude on this path for the duration of the rollout. Compounding it, old pods still run the unconditional-`DEL`-on-failed-acquire behavior this PR fixes, so they can delete new pods' token locks on any key. Since this call site is the one place where the Redis lock is the sole concurrency guard (no DB row lock — see my other comment), the deploy window is exactly where a race could materialize.

No code change needed — suggest a line in the release notes: deploy while prepaid-assignment traffic is idle, or drain old workers before starting the new fleet.

app/Services/Utils/LockManagerService.php
```diff
@@ -41,71 +42,71 @@ public function __construct(ICacheService $cache_service){
...
+     * @param string $token ownership token returned by acquireLock
      */
-    public function releaseLock(string $name):LockManagerService
+    public function releaseLock(string $name, string $token): void
     {
-        Log::debug(sprintf("LockManagerService::releaseLock name %s",$name));
-        $this->cache_service->delete($name);
-        return $this;
+        Log::debug(sprintf("LockManagerService::releaseLock name %s", $name));
+        $this->cache_service->deleteIfValueMatches($name, $token);
```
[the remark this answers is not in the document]
smarcet: @romanetar Building on this suggestion: logging the discarded `false` is the right baseline, but a mismatch detected at release time is *actionable*, not just observable — consider escalating it to a typed exception so the caller can roll back.

Every `lock()` call site in `SummitOrderService` runs inside `tx_service->transaction`, and the release happens in `lock()`'s `finally` before the transaction closure returns. Throwing on a genuine ownership mismatch therefore propagates out of the closure and triggers a DB rollback: the holder that lost its lease aborts instead of committing on top of a competitor's work. That converts silent exclusion loss into "lease-loser aborts" semantics — effectively fencing at commit time, which is the strongest guarantee a single-instance Redlock can offer without real fencing tokens in the storage layer.

Four conditions to do it safely:

1. **Tri-state release result, not bool.** `deleteIfValueMatches()` returns `false` in three distinct situations: lease genuinely lost (key gone or value mismatch), Redis unreachable (`retryOnConnectionError` falls back to `false`), and a retried Lua call whose first response was lost after a successful delete. Throwing on all three would turn a Redis blip into a rollback of every in-flight locked operation — an availability regression. The cache layer should report `RELEASED` / `NOT_OWNED` / `UNKNOWN`, and `lock()` should throw only on `NOT_OWNED` and log `UNKNOWN`.
2. **Never mask the callback's own exception.** The release runs in `finally`; if the callback already threw, a throw from `finally` replaces it as the visible exception type (PHP chains the original as `previous`, but upstream handlers see the new type). Only throw when the callback completed successfully; otherwise just log.
3. **Move non-transactional side effects out of the lock first.** `SendAttendeeInvitationEmail::dispatch` currently runs inside the `AutoAssignPrePaidTicketTask` callback, and its queue write does not participate in the DB transaction — a rollback would leave a scheduled email pointing at an assignment that no longer exists. Item 1 of the PR description (dispatch after the lock is released) is a prerequisite for this design, not just a follow-up.
4. **Type the exception** (e.g. `LockLostException`) so the saga tasks — which already have `undo()` compensation paths — can distinguish it from validation failures instead of catching a generic `RuntimeException`.

Combined with the pessimistic row lock suggested in my comment on `AutoAssignPrePaidTicketTask`, this becomes defense in depth: the DB lock prevents the race, the release check detects anything that slips through.

smarcet (changes requested):
@romanetar please re review

app/Services/Model/Imp/SummitOrderService.php
romanetar: Verified both points against the code and applied fixes for both (will push shortly).

**DB pessimistic lock for `AutoAssignPrePaidTicketTask`**: injected `ISummitRegistrationPromoCodeRepository` into the task (wired from `SagaFactory::$promo_code_repository`, which was already available there) and replaced the unlocked `$this->summit->getPromoCodeByCode($promo_code_val)` with `$this->promo_code_repository->getByValueExclusiveLock($this->summit, $promo_code_val)` — same pattern as `ApplyPromoCodeTask`. That row lock (`PESSIMISTIC_WRITE` + `HINT_REFRESH`) now covers the whole check-then-act sequence (`hasPrePaidTicketsAssignedBy` → `getNextAvailableTicketPerType` → `addTicket`) and is held through the enclosing transaction's commit, closing both the TTL-expiry hole and the release-before-commit window. The Redis lock stays in place as the fail-fast layer.

**`releaseLock()` observability**: it now checks the `deleteIfValueMatches()` result and, on a mismatch, logs a warning (lock name + token) and increments a `lock_manager.release_mismatch` counter via the existing `ICacheService::incCounter()` — no new metrics infra introduced. Added a regression test asserting the counter fires when release fails.

Ran `tests/SummitOrderServiceTest.php` against the real DB/Redis containers — no regressions (the 3 prepaid-ticket tests are pre-existing `markTestSkipped('broken test.')`, unrelated to this change, left as-is/out of scope for this PR).

app/Services/Utils/RedisCacheService.php
romanetar: Confirmed and fixed. Reproduced the exact failure against the real Predis-backed Redis container before the fix — `Array to string conversion` warning followed by `Predis\Response\ServerException: ERR syntax error` on every call, uncaught by `retryOnConnectionError` since it only catches `PredisConnectionException|\RedisException`, exactly as described.

Applied your suggested fix, mirroring `addSingleValue`'s variadic form:

```php
public function incCounter($counter_name, $ttl = 0)
{
    return $this->retryOnConnectionError(function ($conn) use ($counter_name, $ttl) {
        if ($ttl > 0) {
            if ($conn->set($counter_name, 1, 'EX', (int)$ttl, 'NX') !== null) return 1;
        } else {
            if ($conn->set($counter_name, 1, 'NX') !== null) return 1;
        }
        return (int)$conn->incr($counter_name);
    }, 0);
}
```

Verified against the real container: first call sets the key (returns 1), subsequent calls hit the `NX` failure path and fall through to `incr` (2, 3, ...), both with and without a TTL. No exceptions.

Kept `incCounter` rather than reverting the hunk, since it's now the target of the `lock_manager.release_mismatch` counter added in `LockManagerService::releaseLock` for the other thread on this PR — so it does have a caller as of this branch.

tests/Integration/RedisCacheServiceAddSingleValueTest.php
romanetar: Both done.

**CI wiring**: added `{ name: "Integration", filter: "tests/Integration/" }` to the `integration-tests` matrix in `.github/workflows/push.yml`, right after the `Services` entry. Redis is already provisioned there via `supercharge/redis-github-action`, so no other setup was needed. Verified this is the correct invocation form (bare directory path, matching the `OAuth2Tests`/`EntityModelUnitTests`/`AuditUnitTests` entries) — `vendor/bin/phpunit --filter 'tests/Integration/'` actually matches nothing and prints "No tests executed!" since `--filter` runs a regex against test/class names, not paths, so that form would have silently run zero tests in CI.

**`deleteIfValueMatches` coverage**: extended `RedisCacheServiceAddSingleValueTest` with two cases — matching token deletes the key, non-matching token leaves it intact (asserted both the return value and the key's survival/state). Ran all 4 tests in the file against the real Redis container: pass.

app/Services/Model/Imp/SummitOrderService.php
romanetar: Good flag, and confirmed against `main`'s `LockManagerService` — `releaseLock()` there is an unconditional `delete($name)` and `lock()`'s `finally` calls it even after `acquireLock` throws `UnacquiredLockException`, so an old pod that loses contention on any key will delete it out from under whoever holds it, no ownership check. Combined with the key-namespace split you describe (missing dot on old pods vs. present on new), the two fleets neither mutually exclude on this path nor leave each other's locks alone during the rollout — exactly the window you're calling out.

No code change from this thread, but worth noting: the DB pessimistic lock added for your other comment on this same call site (`getByValueExclusiveLock` on the promo code, `PESSIMISTIC_WRITE` + `HINT_REFRESH`, held through commit) closes the concrete failure mode here — old and new pods share the same MySQL row lock regardless of what happens to the Redis key naming or unconditional deletes on either side. So the double-assignment scenario this thread worried about is now covered even mid-rollout; Redis staying inconsistent across fleets during the deploy window would at most cause spurious contention/retries, not an actual double-booked prepaid ticket.

Will add the release-note line about draining old workers / deploying during idle prepaid-assignment traffic anyway, since the underlying old-pod unconditional-release bug can still stomp on other locks in the system during the transition, not just this one.

.github/workflows/push.yml
```diff
@@ -63,12 +63,13 @@ jobs:
                     - { name: "AuditOtlpStrategyTest", filter: "--filter AuditOtlpStrategyTest" }
                     - { name: "AuditEventTypesTest", filter: "--filter AuditEventTypesTest" }
                     - { name: "GuzzleTracingTest", filter: "--filter GuzzleTracingTest" }
-                    - { name: "Repositories", filter: "tests/Repositories/" }
-                    - { name: "Services", filter: "tests/Unit/Services/" }
                     - { name: "CacheOptimizations", filter: "--filter '(PresentationSpeakerCacheTest|ResourceServerContextTest)'" }
                     # Named by path because no job in this matrix runs the tests/ root, only its
                     # subdirectories - a file added there runs nowhere unless it is listed here.
                     - { name: "PresentationMediaUploads", filter: "tests/PresentationMediaUploadsTest.php tests/PresentationMediaUploadsVisibilityTest.php tests/PresentationSerializerCacheKeyTest.php" }
+                    - { name: "Repositories", filter: "--filter tests/Repositories/" }
```
smarcet: @romanetar this is a CI coverage regression. These entries changed from PHPUnit path arguments to `--filter tests/...`, but PHPUnit `--filter` matches test/class names, not directories. As a result, the Services and Repositories jobs now execute zero tests while still exiting 0.

I reproduced it with:
`vendor/bin/phpunit --filter tests/Unit/Services/ --do-not-cache-result`
`vendor/bin/phpunit --filter tests/Repositories/ --do-not-cache-result`

Both print `No tests executed!` and exit 0. Please keep `tests/Unit/Services/` and `tests/Repositories/` as path arguments, and add `tests/Integration/` as a path argument too.

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -1539,7 +1548,7 @@ public function run(array $formerState): array
...
                 function () use ($promo_code_val, $type_id) {
```
smarcet: @romanetar **Missing `$ticket_dto` in the closure's `use()` clause — attendee-submitted data silently dropped.**

The closure at this line (`function () use ($promo_code_val, $type_id) { ... }`) reads `$ticket_dto['attendee_company']`, `['attendee_first_name']`, and `['attendee_last_name']` in its body, but `$ticket_dto` isn't captured. PHP nested closures don't inherit an enclosing closure's locals automatically — only `use()`'d variables (and `$this`, which is why `$this->owner`/`$this->payload` work here without a `use` entry). Inside this closure `$ticket_dto` is undefined, so every one of these reads silently falls through the `??`/`empty()` fallback to `$this->payload['owner_*']` / `$this->owner->get*()` — the order owner's own profile, not the attendee actually being assigned the ticket.

I confirmed this against the branch directly: a regression test that submits `attendee_company = "Attendee Co"` on the ticket while the order owner's company is `"Tipit"` gets back an attendee with company `"Tipit"` — the ticket-level value never reaches `SummitAttendeeFactory`. This is CodeRabbit's earlier finding on this same line (`cr-comment:v1:4850e25b`); it looks like it slipped through unaddressed while the other three review threads on this file were fixed.

Suggested fix:
```diff
 $order = $this->lock_service->lock('ticket_type.' . $type_id . '.promo_code.' . $promo_code_val . '.sell.lock',
-    function () use ($promo_code_val, $type_id) {
+    function () use ($promo_code_val, $type_id, $ticket_dto) {
```

I verified this one-line change is sufficient — with it applied, the regression test below passes (attendee gets the ticket-level company/first/last name), and fails without it (`Expected 'Attendee Co' / Actual 'Tipit'`), so it's a genuine RED/GREEN pair, not a fixture artifact:

```php
public function testAutoAssignPrePaidTicketUsesTicketLevelAttendeeData() {

    // Fixture registration window starts tomorrow (see InsertSummitTestData);
    // open it now so this test isn't blocked by an unrelated precondition.
    self::$summit->setRegistrationBeginDate(new \DateTime('-1 day'));
    self::$summit->setRegistrationEndDate(new \DateTime('+30 days'));
    self::$em->persist(self::$summit);
    self::$em->flush();

    // Build a dedicated unassigned, paid, offline ticket so this test does not
    // depend on the base fixture's order #0 (already-assigned attendee / online
    // payment method, neither of which qualifies for prepaid pickup).
    $owner = self::$defaultMember;

    $order = new SummitOrder();
    $order->setSummit(self::$summit);
    $order->setOwner($owner);
    $order->setPaymentMethodOffline();
    $order->generateNumber();

    $ticket = new SummitAttendeeTicket();
    $ticket->setTicketType(self::$default_ticket_type);
    $order->addTicket($ticket);
    $ticket->activate();
    $ticket->generateNumber();
    $ticket->generateQRCode();

    self::$summit->addOrder($order);
    self::$em->persist($order);
    self::$em->flush();

    $order->setPaid();

    self::$default_prepaid_discount_code->clearTickets();
    self::$default_prepaid_discount_code->addTicket($ticket);
    self::$em->persist(self::$default_prepaid_discount_code);
    self::$em->persist($order);
    self::$em->flush();

    $service = App::make(ISummitOrderService::class);

    $payload = [
        "owner_email"      => $owner->getEmail(),
        "owner_first_name" => $owner->getFirstName(),
        "owner_last_name"  => $owner->getLastName(),
        "owner_company"    => $owner->getCompany(),
        "tickets" => [
            [
                "type_id"             => self::$default_ticket_type->getId(),
                "promo_code"          => self::$default_prepaid_discount_code->getCode(),
                // Attendee is a different person than the order owner -
                // this is the scenario AutoAssignPrePaidTicketTask exists for.
                "attendee_company"    => "Attendee Co",
                "attendee_first_name" => "Jane",
                "attendee_last_name"  => "Doe",
            ],
        ]
    ];

    $result_order = $service->reserve($owner, self::$summit, $payload);
    $attendee     = $result_order->getTickets()->first()->getOwner();

    $this->assertEquals("Attendee Co", $attendee->getCompanyName());
    $this->assertEquals("Jane", $attendee->getFirstName());
    $this->assertEquals("Doe", $attendee->getSurname());
}
```

It doesn't reuse the base fixture's order #0 or touch the three pre-existing `markTestSkipped('broken test.')` tests in this file — those are broken for unrelated reasons (a hardcoded `summit_repository->find(3800)` lookup, and the base fixture's registration window starting tomorrow rather than today).

app/Services/Utils/RedisCacheService.php
```diff
@@ -306,12 +307,11 @@ public function setSingleValue($key, $value, $ttl = 0)
     public function addSingleValue($key, $value, $ttl = 0)
     {
         return $this->retryOnConnectionError(function ($conn) use ($key, $value, $ttl) {
-            $res = $conn->setnx($key, $value);
-            if ($res && $ttl > 0) {
-                $conn->expire($key, $ttl);
+            if ($ttl > 0) {
+                return $conn->set($key, $value, 'EX', (int)$ttl, 'NX') !== null;
```
smarcet: @romanetar **`!== null` to detect a `SET ... NX` miss is Predis-specific — under PhpRedis it silently defeats the lock's mutual exclusion.**

`addSingleValue()` here and `incCounter()` (`RedisCacheService.php:243`) both check `$conn->set(..., 'NX') !== null` to detect whether the `SET` acquired the key. That's correct for Predis, which returns `null` on an NX-miss — but PhpRedis (`ext-redis`, the C extension) returns `false` on an NX-miss for this same variadic form, and `false !== null` evaluates to `true`. So under `REDIS_CLIENT=phpredis`, a second `addSingleValue()` call on a key that's still held reports success instead of failure.

Concretely: `LockManagerService::acquireLock()` is built entirely on this return value — a second caller racing for an already-held lock would get back a token and believe it holds the lock, while the first token is still the actual owner. That's the exact mutual-exclusion guarantee this PR exists to add. `incCounter()` degrades the same way: the `lock_manager.release_mismatch` counter this PR just wired up would silently stick at 1 instead of incrementing.

This isn't a new observation — it's the first point of an existing, still-open review comment on this same line (driver-compatibility, "If the driver is ever switched to PhpRedis, `set()` returns `false` on an NX-miss (not `null`), silently breaking the `!== null` check"). The reply that followed fixed the CI wiring and added `deleteIfValueMatches` coverage but didn't touch this point, so it's been carried unresolved since.

Verified directly: `docker exec -e REDIS_CLIENT=phpredis summit-api vendor/bin/phpunit tests/Integration/RedisCacheServiceAddSingleValueTest.php` fails at the NX-reacquire assertion (`Failed asserting that true is false`), and a two-call `incCounter($key, 30)` probe returns `[1, 1, "1", 30]` under PhpRedis vs. the correct `[1, 2, "2", 30]` under Predis.

One mitigating note: I checked `argocd-apps/summit-api`'s `values.yaml` / `values-prod.yaml` / `values-stage.yaml` — `REDIS_CLIENT` isn't set in either, so both currently run on the `predis` default and aren't hitting this today. But the `redis` (PhpRedis) extension is already compiled into the app image, nothing in config or CI guards against someone setting the env var, and CI never runs this suite under PhpRedis either — so it's a live landmine with zero coverage, not a hypothetical one.

Suggested fix — treat both driver's failure sentinels the same way, consistent for `addSingleValue` and `incCounter`:
```diff
-                if ($conn->set($counter_name, 1, 'EX', (int)$ttl, 'NX') !== null) return 1;
+                if ($this->setNxSucceeded($conn->set($counter_name, 1, 'EX', (int)$ttl, 'NX'))) return 1;
...
-                if ($conn->set($counter_name, 1, 'NX') !== null) return 1;
+                if ($this->setNxSucceeded($conn->set($counter_name, 1, 'NX'))) return 1;
...
-            return $conn->set($key, $value, 'EX', (int)$ttl, 'NX') !== null;
+            return $this->setNxSucceeded($conn->set($key, $value, 'EX', (int)$ttl, 'NX'));
...
-            return $conn->set($key, $value, 'NX') !== null;
+            return $this->setNxSucceeded($conn->set($key, $value, 'NX'));
```
with a small private helper `setNxSucceeded($result): bool { return $result !== null && $result !== false; }` (any real success value — a `Predis\Response\Status` object or PhpRedis's `true`/`1` — is truthy against both checks; only the two drivers' respective failure sentinels are excluded). Worth also adding a `deleteIfValueMatches`-style `@group integration` case that runs under both `REDIS_CLIENT=predis` and `REDIS_CLIENT=phpredis` in CI, so this class of bug can't reappear silently again.

tests/Unit/Services/LockManagerServiceOwnershipTest.php
```diff
@@ -0,0 +1,164 @@
...
+     * Structural assertion: addSingleValue is called exactly once per
+     * acquisition attempt (not two separate calls for setnx + expire).
+     * The call must carry the lock name, a string token, and the lifetime.
+     */
+    public function testAddSingleValueCalledOnceWithTokenAndLifetime(): void
+    {
+        $cache = Mockery::mock(ICacheService::class);
+        $cache->shouldReceive('addSingleValue')
+              ->once()
+              ->with('test.lock', Mockery::type('string'), 3600)
```
smarcet: @romanetar **Token identity isn't asserted between `addSingleValue` and `deleteIfValueMatches` — the one guarantee this whole PR adds.**

`Mockery::type('string')` on line 122 accepts any string, and `deleteIfValueMatches` on line 124 has no `with()` constraint at all. So this test would still pass if `releaseLock()` released using a hardcoded literal, a stale token from a previous call, or a completely wrong value — as long as it's *some* string. The one thing this PR is actually about (the release only succeeds when the token matches the one `acquireLock` handed out) is exactly what's left unverified here.

This has been flagged twice already by CodeRabbit on earlier revisions of this same test (once as "assert the token is non-empty," once more precisely as "assert release uses the exact acquired token," with a ready-to-use diff) and both times it was left unaddressed while the file kept changing around it.

To be clear, this isn't a live bug — `LockManagerService::lock()` trivially threads the same `$token` variable from `acquireLock()` into `releaseLock()`, so nothing is broken today. But it means a future refactor that breaks that threading (e.g. re-deriving the token instead of reusing it, or introducing the mutable-instance-state pattern flagged elsewhere on this PR) would sail through this suite undetected.

Suggested fix — capture the token on the `addSingleValue` expectation and assert `deleteIfValueMatches` receives the same value:
```diff
     public function testAddSingleValueCalledOnceWithTokenAndLifetime(): void
     {
         $cache = Mockery::mock(ICacheService::class);
+        $token = null;
         $cache->shouldReceive('addSingleValue')
               ->once()
-              ->with('test.lock', Mockery::type('string'), 3600)
+              ->with(
+                  'test.lock',
+                  Mockery::on(function ($value) use (&$token) {
+                      if (!is_string($value) || $value === '') return false;
+                      $token = $value;
+                      return true;
+                  }),
+                  3600
+              )
               ->andReturn(true);
-        $cache->shouldReceive('deleteIfValueMatches')->once()->andReturn(true);
+        $cache->shouldReceive('deleteIfValueMatches')
+              ->once()
+              ->with('test.lock', Mockery::on(fn($value) => $value === $token))
+              ->andReturn(true);
```

smarcet (changes requested):
@romanetar please re review i do still see some pending issues

app/Services/Utils/RedisCacheService.php
smarcet: Reopening — this thread has two distinct asks and only one was addressed.

Point 2 (atomicity / TTL-in-one-command integration coverage) was fixed and is covered by the `Integration` test suite added to this PR. But point 1 (driver compatibility — "If the driver is ever switched to PhpRedis, `set()` returns `false` on an NX-miss (not `null`), silently breaking the `!== null` check") was never touched, and this got marked resolved anyway.

I verified point 1 is a real, currently-live bug in the code as it stands today: under `REDIS_CLIENT=phpredis`, `addSingleValue()`/`incCounter()` both report success on an NX-miss, which lets a second `acquireLock()` succeed while the first token still owns the lock. Full writeup, reproduction, and suggested fix are in a new comment on `RedisCacheService.php:311` (current line for `addSingleValue`) rather than here, since the diff context on this line has moved.

@romanetar — see that comment for the fix; leaving this one open until it's addressed there.

.github/workflows/push.yml
romanetar: Good catch, fixed in 0795ca9b9 — reverted `Repositories` and `Services` back to path arguments.

app/Services/Model/Imp/SummitOrderService.php
romanetar: Confirmed and fixed in 685d70bd4 — added `$ticket_dto` to the closure's `use()` clause and added your regression test (`testAutoAssignPrePaidTicketUsesTicketLevelAttendeeData`). Verified locally: RED (`Expected 'Attendee Co' / Actual 'Tipit'`) with the closure as-is, GREEN with the one-line fix. Full `SummitOrderServiceTest` suite passes (22 tests, 3 unrelated pre-existing skips).

app/Services/Utils/RedisCacheService.php
romanetar: Confirmed and fixed in 126169d1e. Added a shared `setNxSucceeded()` helper that excludes both drivers' failure sentinels (`null` for Predis, `false` for PhpRedis) and used it in both `addSingleValue` and `incCounter`.

Verified locally against both drivers:
- Before the fix: `REDIS_CLIENT=phpredis` — `addSingleValue`'s NX-reacquire test fails (`Failed asserting that true is false`), and a two-call `incCounter` probe returns `1, 1` instead of `1, 2`. Matches your repro exactly.
- After the fix: passes under both `REDIS_CLIENT=predis` and `REDIS_CLIENT=phpredis`.

Also added an `incCounter` regression case to `RedisCacheServiceAddSingleValueTest` and a new `IntegrationPhpRedis` CI matrix entry that runs `tests/Integration/` with `REDIS_CLIENT=phpredis`, alongside the existing predis-default run, so this can't regress silently again.

tests/Unit/Services/LockManagerServiceOwnershipTest.php
romanetar: Confirmed and fixed in 1f5bae739. One catch on the suggested diff itself: `Mockery::on(fn($value) => $value === $token)` doesn't work as written — arrow functions capture $token **by value** at closure-creation time, which is before `addSingleValue`'s callback has run, so $token is still `null` at that point. Confirmed this by applying the diff verbatim: it throws `NoMatchingExpectationException` on every run, not just on a real regression.

Used a regular closure with `use (&$token)` instead so it reads the current value at call time. Then verified the test is actually meaningful (not just passing-by-luck) by injecting a token-threading regression into `LockManagerService::lock()` (hardcoding the token passed to `releaseLock`) — the test failed as expected, then reverting the injected bug brings it back to green.

app/Services/Utils/RedisCacheService.php
romanetar: This is the same driver-compatibility bug flagged on RedisCacheService.php:311 — fixed there in 126169d1e (shared `setNxSucceeded()` helper covering both Predis's `null` and PhpRedis's `false` NX-miss sentinels, used in both `addSingleValue` and `incCounter`), verified under both `REDIS_CLIENT=predis` and `REDIS_CLIENT=phpredis`, plus the new `IntegrationPhpRedis` CI job. See my reply on that thread for the full verification. Closing this one out since it's covered by the same fix.

smarcet (approved):
LGTM

PR #542, opened 2026-05-12
fix: disable metric formatters

andrestejerina97:
ref: https://app.clickup.com/t/86b9pvafe

## Summary by CodeRabbit

* **New Features**
  * Per-entity audit disabling via configuration.

* **Improvements**
  * Audit pipeline now checks per-entity disable flags earlier and skips processing when disabled.
  * Public API extended so callers can query whether auditing is disabled for a given subject.

* **Tests**
  * Added tests validating per-entity enable/disable behavior and that audit processing returns early when disabled.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenStackweb/summit-api/pull/542?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/Audit/AuditLogFormatterFactory.php
+++ b/app/Audit/AuditLogFormatterFactory.php
@@ -22,7 +22,6 @@ use App\Audit\ConcreteFormatters\DefaultEntityManyToManyCollectionUpdateAuditLog
 use App\Audit\ConcreteFormatters\DefaultEntityManyToManyCollectionDeleteAuditLogFormatter;
 use App\Audit\ConcreteFormatters\EntityUpdateAuditLogFormatter;
 use App\Audit\Interfaces\IAuditStrategy;
-use Doctrine\Common\Util\ClassUtils;
 use Doctrine\ORM\PersistentCollection;
 use Illuminate\Support\Facades\Log;
 use Doctrine\ORM\Mapping\ClassMetadata;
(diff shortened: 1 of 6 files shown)


caseylocker (approved):
LGTM. Nothing blocking. Follows the ticket direction.

app/Audit/AuditLogFormatterFactory.php
```diff
@@ -190,4 +203,32 @@ private function routeMatches(string $route, string $actual_route): bool
...
+
+    private function getSubjectClass(mixed $subject): ?string
+    {
+        if (!is_object($subject)) {
+            return null;
+        }
+
+        if (class_exists(ClassUtils::class)) {
+            return ClassUtils::getClass($subject);
+        }
```
smarcet: @andrestejerina97 
**Dead code : `ClassUtils` isn't available in this project.**

This project runs `doctrine/orm` 3.3.3 with no `doctrine/common` package, so `Doctrine\Common\Util\ClassUtils` doesn't exist anywhere in `vendor/`. `class_exists(ClassUtils::class)` is therefore always `false`, this branch never executes, and `getSubjectClass()` always falls through to `get_class()` on line 232. The `use Doctrine\Common\Util\ClassUtils;` import (line 25) is unused at runtime.

This is harmless for the metric-disable goal — the audit listener passes real entity instances from the UnitOfWork (not lazy proxies), and `get_class()` already matched these entities before this PR. But the code reads as if Doctrine proxies are handled when they aren't, which is misleading for the next reader.

Suggest one of:
- Remove the `ClassUtils` import and this branch, keeping just `get_class($subject)`; or
- If proxy-safe resolution is actually intended, do it the ORM 3 way, e.g. `$em->getClassMetadata(get_class($subject))->getName()` or a `Doctrine\Persistence\Proxy` instanceof check.

smarcet (approved):
LGTM

PR #543, opened 2026-05-12
feat(speakers): add unique activities count endpoints for speakers and submitters

mulldug:
ref: https://app.clickup.com/t/86b9b1qrk

- Add GET /summits/{id}/speakers/all/events/count and GET /summits/{id}/submitters/all/events/count controller methods with OpenAPI attributes and route registrations
- Implement getUniqueActivitiesCountBySummit in DoctrineSpeakerRepository and DoctrineMemberRepository using two-stage DQL→raw SQL COUNT(DISTINCT)
- Add interface declarations to ISpeakerRepository and IMemberRepository
- Register both endpoints in ApiEndpointsSeeder with ReadSummitData scopes
- Add PHPUnit tests for both HTTP endpoints and repository-level method
- Fix null-guard bug in DoctrineMemberRepository::applyExtraJoins

## Summary by CodeRabbit

* **New Features**
  * Two new summit APIs returning unique activity counts for speakers and submitters (GET .../speakers/all/events/count and .../submitters/all/events/count) with optional filters.

* **Bug Fixes**
  * Speaker/submitter boolean-style filters (has_*_presentations) are now optional when provided as true/false.
  * Query deduplication improvements in activity counting.

* **Tests**
  * Added API and repository tests validating activity counts and filter behavior.

* **Chores**
  * Registered endpoints via seeder and migration; added route entries.
  * Updated example env quoting and .gitignore to exclude local docker override.


app/Http/Middleware/OAuth2BearerAccessTokenRequestValidator.php
```diff
@@ -173,7 +173,7 @@ public function handle($request, Closure $next)
             }
             if (
                 $token_info->getApplicationType() === 'JS_CLIENT'
-                && (is_null($origin) || empty($origin)|| str_contains($token_info->getAllowedOrigins(), $origin) === false )
+                && (is_null($origin) || empty($origin)|| str_contains($token_info->getAllowedOrigins(), rtrim($origin, '/')) === false )
             ) {
```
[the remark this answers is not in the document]
mulldug: This was an existing issue not created by any work related to this PR.  The fix has been implemented in a separate PR.

database/seeders/ConfigSeeder.php
```diff
@@ -31,6 +31,16 @@ public function run()
...
             $em->clear();
+
+            // evict the L2 (second-level) cache for all resource-server entities so that
+            // stale cached IDs from previous seeder runs don't cause FK violations after
+            // the raw SQL DELETEs below reset the auto-increment sequence.
+            $l2Cache = $em->getCache();
+            if ($l2Cache !== null) {
+                $l2Cache->evictEntityRegions();
+                $l2Cache->evictQueryRegions();
+            }
```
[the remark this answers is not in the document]
mulldug: Both evictEntityRegions() and evictQueryRegions() exist in my installed version (3.3.3) of Doctrine ORM (DefaultCache.php lines 103 and 193).  The method signatures can be found in vendor/doctrine/orm/src/Cache/DefaultCache.php.  The containers were destroyed and recreated repeatedly as part of testing and no issues were reported.  Is it possible this issue is caused by differing versions?  If so maybe the version constraints should be tightened.  Otherwise the code would need to handle both APIs.

app/Repositories/Summit/DoctrineMemberRepository.php
```diff
@@ -638,6 +638,56 @@ function ($query) {
...
+INNER JOIN SummitEvent se ON se.ID = p.ID
+WHERE se.SummitID = :summit_id
+AND p.CreatedByID IN (:member_ids)
+SQL;
+
+        $stm = $this->getEntityManager()->getConnection()->executeQuery(
+            $sql,
+            ['summit_id' => $summit->getId(), 'member_ids' => $memberIds],
+            ['summit_id' => \Doctrine\DBAL\ParameterType::INTEGER, 'member_ids' => \Doctrine\DBAL\ArrayParameterType::INTEGER]
+        );
```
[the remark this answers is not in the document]
mulldug: This has been fixed.

app/Http/Middleware/OAuth2BearerAccessTokenRequestValidator.php
```diff
@@ -173,7 +173,7 @@ public function handle($request, Closure $next)
             }
             if (
                 $token_info->getApplicationType() === 'JS_CLIENT'
-                && (is_null($origin) || empty($origin)|| str_contains($token_info->getAllowedOrigins(), $origin) === false )
+                && (is_null($origin) || empty($origin)|| str_contains($token_info->getAllowedOrigins(), rtrim($origin, '/')) === false )
```
smarcet: @mulldug, please remove this change from this PR. This kind of change should be submitted in a separate PR with the appropriate rationale.

database/seeders/ConfigSeeder.php
```diff
@@ -31,6 +31,16 @@ public function run()
             // clear all
             $em = Registry::getManager(ResourceServerEntity::EntityManager);
             $em->clear();
+
```
smarcet: @mulldug, please remove this change from this PR. This kind of change should be submitted in a separate PR with the appropriate rationale.

docker-compose.override.testing.yml
```diff
@@ -0,0 +1,16 @@
+# This file is used to run this project locally alongside openstackid without port conflicts.
```
smarcet: @mulldug, please remove this change from this PR. This kind of change should be submitted in a separate PR with the appropriate rationale.

database/seeders/ApiEndpointsSeeder.php
```diff
@@ -3733,6 +3733,20 @@ private function seedSummitEndpoints()
                     IGroup::SummitRegistrationAdmins
                 ]
             ],
+            [
+                'name' => 'get-submitters-activities-count',
```
smarcet, line 3737: @mulldug we do need also to create the proper migrations to insert this 
please do check this 
https://github.com/OpenStackweb/summit-api/blob/main/database/migrations/config/Version20260412000000.php

app/ModelSerializers/Summit/SummitSerializer.php
```diff
@@ -320,24 +320,26 @@ public function serialize($expand = null, array $fields = [], array $relations =
...
             ) {
-
-                $values['payment_profiles'][] =
-                    SerializerRegistry::getInstance()->getSerializer
-                    (
-                        $build_default_payment_gateway_profile_strategy->build(IPaymentConstants::ApplicationTypeRegistration),
-                        $this->getSerializerType()
-                    )->serialize(AbstractSerializer::filterExpandByPrefix($expand, 'payment_profiles'));
-
+                try {
```
smarcet: @mulldug, please remove this change from this PR. This kind of change should be submitted in a separate PR with the appropriate rationale.

smarcet:
## Test coverage for the new activities-count endpoints

The new count endpoints are exercised, but the assertions are happy-path only and one repository test is effectively a no-op in CI. Worth tightening before merge:

- The API tests (`OAuth2SummitSpeakersApiTest`, `OAuth2SummitSubmittersApiTest`) only assert `count > 0` / `count >= 0` and `filtered <= unfiltered`. A dedup bug, double-counting of moderators, or a wrong join would still pass. Please add at least one assertion against an **exact expected count** on a known fixture — ideally a presentation with multiple assigned speakers plus a moderator, so the speaker/moderator union + `array_unique` dedup is actually verified.
- `SubmitterRepositoryTest::testGetUniqueActivitiesCountBySummit` calls `markTestSkipped` when summit 3401 is absent, so it provides **zero coverage on a clean CI database**. Please drive it off the seeded test summit (as the API tests do) so the repository method actually runs in CI.

See the inline comments for the specific spots.

tests/SubmitterRepositoryTest.php
```diff
@@ -76,4 +78,26 @@ public function testGetSubmittersIdsBySummit(){
 
         self::assertNotEmpty($submitterIds);
     }
+
+    public function testGetUniqueActivitiesCountBySummit(){
+        $submitter_repository = EntityManager::getRepository(Member::class);
+        $summit_repository = EntityManager::getRepository(Summit::class);
+
+        $summit = $summit_repository->find(3401);
+        if (is_null($summit)) $this->markTestSkipped('Summit 3401 not in test DB');
```
smarcet: This test no-ops on a clean CI database: summit 3401 won't exist, so it hits `markTestSkipped` and provides zero coverage for `getUniqueActivitiesCountBySummit`. Please drive it off the seeded test summit (as the API tests do) so the repository method is actually exercised in CI rather than skipped.

tests/oauth2/OAuth2SummitSpeakersApiTest.php
```diff
@@ -1871,4 +1871,100 @@ public function testDeclineSpeakerEditPermission()
...
+            [],
+            $headers
+        );
+
+        $content = $response->getContent();
+        $this->assertResponseStatus(200);
+        $data = json_decode($content);
+        $this->assertNotNull($data);
+        $this->assertTrue(isset($data->count));
+        $this->assertGreaterThan(0, $data->count);
```
smarcet: Happy-path only: `count > 0` would still pass if moderators are double-counted, the `array_unique` dedup is broken, or the join is wrong. Please assert an **exact expected count** against a known fixture — ideally a presentation with multiple assigned speakers plus a moderator — so the speaker/moderator union + dedup is actually verified, not just "non-zero".

app/Repositories/Summit/DoctrineSpeakerRepository.php
```diff
@@ -697,6 +697,53 @@ function ($query) {
...
+            ->where('p.summit = :summit')
+            ->setParameter('summit', $summit);
+
+        if (!is_null($filter)) {
+            $filter->apply2Query($moderatorQb, $this->getFilterMappings($filter));
+        }
+
+        // Take the union of both result sets and count distinct presentation IDs.
+        // Presentation counts per summit are small (<1000) so PHP-side deduplication is fine.
+        $speakerIds   = array_column($speakerQb->getQuery()->getScalarResult(), 'id');
```
smarcet: This count path materializes every `(presentation, speaker)` and `(presentation, moderator)` row into PHP via `getScalarResult()` and dedups with `array_unique`. The "presentation counts per summit are small (<1000)" assumption is unenforced and degrades for large summits, while still executing two full filtered queries.

This is now also inconsistent with `DoctrineMemberRepository::getUniqueActivitiesCountBySummit`, which was already moved to a SQL subquery. Please do the same here — take the union of the two presentation-id sets and `COUNT(DISTINCT p.id)` in DQL/SQL instead of pulling all ids into PHP.

smarcet (changes requested):
@mulldug please review

app/Repositories/Summit/DoctrineSpeakerRepository.php
```diff
@@ -697,6 +697,39 @@ function ($query) {
...
+    public function getUniqueActivitiesCountBySummit(Summit $summit, Filter $filter = null): int
+    {
+        // Single query: cross-join Presentation × PresentationSpeaker, then filter to
+        // rows where the speaker is either an assigned speaker OR the moderator.
+        // COUNT(DISTINCT p.id) deduplicates in SQL — no PHP-side array_unique needed.
+        // All aliases (e, m, rr) are top-level, so getFilterMappings() applies unchanged.
+        $countQb = $this->getEntityManager()->createQueryBuilder()
+            ->select('COUNT(DISTINCT p.id)')
+            ->from('models\summit\Presentation', 'p')
+            ->from('models\summit\PresentationSpeaker', 'e')
```
smarcet, line 716: @mulldug 
This `FROM` clause creates an unbounded cross-join: `PresentationSpeaker` is not scoped to the summit, so the query starts from a product of (presentations in summit) × (every speaker in the entire database) before the EXISTS/moderator clause narrows it down. On a large deployment — say 1K summit presentations and 10K total speakers — the engine evaluates 10M intermediate rows per call.

The member repository avoids this with a two-stage subquery: first collect matching member IDs, then count presentations for those IDs. The same pattern works here — collect matching speaker IDs in a subquery scoped to this summit's assignments, then feed that as an `IN` subquery into the count.

app/Repositories/Summit/DoctrineMemberRepository.php
```diff
@@ -638,6 +638,48 @@ function ($query) {
...
+        if (!is_null($filter)) {
+            $filter->apply2Query($qb, $this->getFilterMappings($filter));
+        }
+
+        // Count distinct presentations using the member query as a subquery — no PHP ID materialization.
+        $countQb = $this->getEntityManager()->createQueryBuilder()
+            ->select("COUNT(DISTINCT p.id)")
+            ->from('models\summit\Presentation', 'p')
+            ->where('p.summit = :summit_outer')
+            ->andWhere("p.created_by IN ({$qb->getDQL()})");
```
smarcet: @mulldug 
Embedding `$qb->getDQL()` via string interpolation is fragile in two ways.

First, the filter mappings for this repository contain hardcoded `:summit` parameter references inside their EXISTS subqueries (`presentations_track_id`, `has_accepted_presentations`, `presentations_title`, etc.). Those parameters are propagated by the copy loop below, but if any future filter mapping introduces a parameter named `:summit_outer` it will silently overwrite the outer query's summit binding and return incorrect counts.

Second, `getDQL()` is an internal Doctrine QueryBuilder API with no stability guarantee on its output format.

Consider using `Expr\In` instead of string interpolation (`$countQb->expr()->in('p.created_by', $qb->getDQL())`) and renaming the inner query's `:summit` parameter to something unambiguous (e.g. `:submitter_summit`) to make the parameter boundary explicit.

smarcet (changes requested):
@mulldug please review.

** Please avoid force-pushing and add new commits instead, so we can easily track the changes. **

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitSpeakersApiController.php
```diff
@@ -392,6 +392,110 @@ function ($page, $per_page, $filter, $order, $applyExtraFilters) use ($summit) {
...
+            ),
+            new OA\Response(response: Response::HTTP_NOT_FOUND, description: 'Not Found'),
+            new OA\Response(response: Response::HTTP_INTERNAL_SERVER_ERROR, description: 'Server Error'),
+        ]
+    )]
+    public function getSpeakersActivitiesCount($summit_id)
+    {
+        return $this->processRequest(function () use ($summit_id) {
+
+            $summit = SummitFinderStrategyFactory::build($this->getRepository(), $this->getResourceServerContext())->find($summit_id);
```
romanetar: @mulldug please cast $summit_id to int

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitSubmittersApiController.php
```diff
@@ -495,4 +495,111 @@ public function send($summit_id)
...
+            new OA\Response(response: Response::HTTP_FORBIDDEN, description: "Forbidden"),
+            new OA\Response(response: Response::HTTP_NOT_FOUND, description: "Summit not found"),
+            new OA\Response(response: Response::HTTP_INTERNAL_SERVER_ERROR, description: "Server Error"),
+        ]
+    )]
+    public function getSubmittersActivitiesCount($summit_id)
+    {
+        return $this->processRequest(function () use ($summit_id) {
+
+            $summit = SummitFinderStrategyFactory::build($this->summit_repository, $this->getResourceServerContext())->find(intval($summit_id));
```
romanetar, line 543: https://github.com/OpenStackweb/summit-api/pull/543/changes#r3357358197

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitSubmittersApiController.php
```diff
@@ -495,4 +495,111 @@ public function send($summit_id)
...
+                $filter->validate([
+                    'id' => 'sometimes|integer',
+                    'not_id' => 'sometimes|integer',
+                    'first_name' => 'sometimes|string',
+                    'last_name' => 'sometimes|string',
+                    'email' => 'sometimes|string',
+                    'full_name' => 'sometimes|string',
+                    'member_id' => 'sometimes|integer',
+                    'member_user_external_id' => 'sometimes|integer',
+                    'has_accepted_presentations' => 'sometimes|string|in:true,false',
```
romanetar, line 584: @mulldug The getSpeakersActivitiesCount validation rules for has_accepted_presentations, has_alternate_presentations, and has_rejected_presentations use sometimes|required|string|in:true,false,
  while the sibling getSubmittersActivitiesCount uses sometimes|string|in:true,false (no required). Since both endpoints accept the same filter semantics, the validation should be consistent
  — pick one form and apply it to both.

romanetar (changes requested):

app/Http/Middleware/OAuth2BearerAccessTokenRequestValidator.php
mulldug: Removed.

database/seeders/ConfigSeeder.php
mulldug: Removed.

docker-compose.override.testing.yml
mulldug: Removed.

database/seeders/ApiEndpointsSeeder.php
mulldug: Added.

app/ModelSerializers/Summit/SummitSerializer.php
mulldug: Removed.

app/Repositories/Summit/DoctrineSpeakerRepository.php
mulldug: The feedback is based on an earlier version of the query. The current implementation does not perform an unbounded scan of PresentationSpeaker — the inner query's WHERE clause constrains it to speakers who belong to the target summit via an EXISTS subquery against PresentationSpeakerAssignment or a moderator join, both filtered by summit_id. Only the matching speaker IDs are returned, and the outer query then counts distinct presentations for those speakers.

Restructuring to start from Presentation (as the member repository does) is not straightforward in the speaker case. The member repository has a single speaker-role relationship to navigate; this method must cover two distinct paths — assigned speaker (PresentationSpeakerAssignment) and moderator. All caller-supplied filters (e.g. has_accepted_presentations, has_alternate_presentations, has_rejected_presentations) are expressed against the e alias on PresentationSpeaker. A Presentation-first structure would require either duplicating every filter mapping for both paths or reintroducing the same inner subquery, which gains nothing over the current design and significantly increases complexity.

If there is a specific query plan concern (e.g. a missing index on PresentationSpeakerAssignment.speaker or Presentation.moderator), that would be worth addressing at the database level. But the query structure itself is not doing a full table scan — the summit predicate is applied in the innermost subquery before any join expansion occurs.

app/Repositories/Summit/DoctrineMemberRepository.php
mulldug: This feedback is based on a previous version.

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitSpeakersApiController.php
```diff
@@ -392,6 +392,110 @@ function ($page, $per_page, $filter, $order, $applyExtraFilters) use ($summit) {
...
+        ],
+        responses: [
+            new OA\Response(
+                response: Response::HTTP_OK,
+                description: 'Unique activities count',
+                content: new OA\JsonContent(
+                    properties: [new OA\Property(property: 'count', type: 'integer')]
+                )
+            ),
+            new OA\Response(response: Response::HTTP_NOT_FOUND, description: 'Not Found'),
```
romanetar: @mulldug, these response codes are inconsistent with those of getSubmittersActivitiesCount. They should have the same response code set.

romanetar (changes requested):
@mulldug please review

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitSpeakersApiController.php
mulldug: getSubmittersActivitiesCount has the correct response codes, so getSpeakersActivitiesCount has been brought into alignment with that method.  They now match up.

smarcet (approved):
LGTM

PR #545, opened 2026-05-13
fix(oauth2): use exact matching for JS client origin validation

mulldug:
ref: https://app.clickup.com/t/86b9xky5x
## Summary

- Replace `str_contains` substring check with `in_array` exact match in `OAuth2BearerAccessTokenRequestValidator` for JS_CLIENT origin validation
- Splits the space-delimited `allowed_origins` string from the IDP introspection response into individual entries, trims each, and strips trailing slashes before comparing
- Prevents crafted short origins (e.g. `example`) from satisfying the guard by matching as a substring of a legitimate allowed origin

## Notes

This is a pre-existing bug, not introduced by recent work. Separated into its own PR to keep the security fix reviewable independently.

## Test plan

- [ ] JS client requests with a valid origin are accepted
- [ ] JS client requests with an origin that is a substring of a valid origin (e.g. `example` when `https://example.com` is allowed) are rejected
- [ ] JS client requests with no origin are rejected

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced OAuth2 Bearer token request validation with stricter origin matching. The validation now requires exact matches instead of substring-based checks, improving security for API requests.

* **Tests**
  * Added comprehensive unit tests for OAuth2 Bearer token request validator covering origin validation scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenStackweb/summit-api/pull/545?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/Http/Middleware/OAuth2BearerAccessTokenRequestValidator.php
+++ b/app/Http/Middleware/OAuth2BearerAccessTokenRequestValidator.php
@@ -171,16 +171,22 @@ class OAuth2BearerAccessTokenRequestValidator
                 );
                 throw new InvalidGrantTypeException(OAuth2Protocol::OAuth2Protocol_Error_InvalidToken);
             }
-            if (
-                $token_info->getApplicationType() === 'JS_CLIENT'
-                && (is_null($origin) || empty($origin)|| str_contains($token_info->getAllowedOrigins(), $origin) === false )
-            ) {
+            if ($token_info->getApplicationType() === 'JS_CLIENT') {
                 //check origins
-                throw new OAuth2ResourceServerException(
-                    403,
-                    OAuth2Protocol::OAuth2Protocol_Error_UnauthorizedClient,
-                    sprintf('invalid origin %s - allowed ones (%s)', $origin, $token_info->getAllowedOrigins())
-                );
+                $allowedOrigins = array_filter(array_map(function ($o) {
+                    $o = trim($o);
+                    if ($o === '') return '';
+                    try { $o = (new Normalizer($o))->normalize(); } catch (\Throwable $e) {}
+                    return rtrim($o, '/');
+                }, explode(' ', $token_info->getAllowedOrigins() ?? '')));
+
+                if (is_null($origin) || empty($origin) || !in_array(rtrim($origin, '/'), $allowedOrigins, true)) {
+                    throw new OAuth2ResourceServerException(
+                        403,
+                        OAuth2Protocol::OAuth2Protocol_Error_UnauthorizedClient,
+                        sprintf('invalid origin %s - allowed ones (%s)', $origin, $token_info->getAllowedOrigins())
+                    );
+                }
             }
             //check scopes
             Log::debug('OAuth2BearerAccessTokenRequestValidator::handle checking token scopes ...');
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ phpunit.xml
 .phpunit.cache/
 docker-compose/mysql/config/*.sql
 docker-compose/mysql/model/*.sql
+docker-compose.override.yml
 package.xml
 .env.dev
 rector.php
--- a/tests/Unit/OAuth2BearerAccessTokenRequestValidatorTest.php
+++ b/tests/Unit/OAuth2BearerAccessTokenRequestValidatorTest.php
@@ -1,0 +1,174 @@
+<?php namespace Tests\Unit;
+/**
+ * Copyright 2026 OpenStack Foundation
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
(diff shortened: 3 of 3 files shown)


app/Http/Middleware/OAuth2BearerAccessTokenRequestValidator.php
```diff
@@ -171,9 +171,13 @@ public function handle($request, Closure $next)
...
                 throw new InvalidGrantTypeException(OAuth2Protocol::OAuth2Protocol_Error_InvalidToken);
             }
+            $allowedOrigins = array_filter(array_map(
+                fn($o) => rtrim(trim($o), '/'),
+                explode(' ', $token_info->getAllowedOrigins() ?? '')
+            ));
             if (
                 $token_info->getApplicationType() === 'JS_CLIENT'
-                && (is_null($origin) || empty($origin)|| str_contains($token_info->getAllowedOrigins(), $origin) === false )
+                && (is_null($origin) || empty($origin) || !in_array(rtrim($origin, '/'), $allowedOrigins, true))
```
smarcet: @mulldug 
**Asymmetric normalization — risk of false 403s for legitimate JS clients**

`$origin` is built by `RequestUtils::getOrigin()`, which runs the header through `URL\Normalizer` (lowercases scheme/host, strips default ports `:80`/`:443`, canonicalizes percent-encoding, resolves dot segments, may add/strip trailing `/`).

`$allowedOrigins` here only gets `trim()` + `rtrim('/')` — no host-lowercasing, no port normalization, no percent-encoding canonicalization.

Combined with strict `in_array(..., true)` (byte-exact, case-sensitive), any IDP record stored as e.g. `https://Example.com`, `https://example.com:443`, or with percent-encoded chars will silently start returning `403 unauthorized_client` after this ships — even though the old `str_contains` accidentally tolerated those mismatches.

Failure mode is silent and uniform across affected clients, and local tests won't catch it (fixtures use `allowed_origins => ''`, so the JS_CLIENT branch is skipped).

**Suggested fix:** normalize each entry through the same `URL\Normalizer` before the strict compare, so both sides come out of the same pipeline:

```php
$allowedOrigins = array_filter(array_map(function ($o) {
    $o = trim($o);
    if ($o === '') return '';
    try { $o = (new \URL\Normalizer($o))->normalize(); } catch (\Throwable $e) {}
    return rtrim($o, '/');
}, explode(' ', $token_info->getAllowedOrigins() ?? '')));
```

**Recommended pre-rollout check:** query the token store for `allowed_origins` entries containing uppercase letters, `:80`/`:443`, or `%`-encoding to surface clients at risk.

smarcet:
@mulldug all proposed PR needs 
1. ref: <ticket_url> header
2. the proper regresion tests that prove the bug:
 Add a unit test with a mocked $token_service returning a token whose allowed_origins = "https://example.com https://foo.bar" and assert (a) example is rejected, (b)             
  https://example.com is accepted, (c) https://example.com/ is accepted, (d) empty Origin is rejected.

smarcet (changes requested):
@mulldug please review comments

app/Http/Middleware/OAuth2BearerAccessTokenRequestValidator.php
```diff
@@ -171,9 +171,13 @@ public function handle($request, Closure $next)
                 );
                 throw new InvalidGrantTypeException(OAuth2Protocol::OAuth2Protocol_Error_InvalidToken);
             }
+            $allowedOrigins = array_filter(array_map(
```
smarcet: @mulldug 
Move the array build inside the JS_CLIENT branch so server-to-server tokens do not pay the cost

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitSubmittersApiController.php
```diff
@@ -495,4 +495,111 @@ public function send($summit_id)
...
+            SummitScopes::ReadAllSummitData,
+        ]]],
+        parameters: [
+            new OA\Parameter(
+                name: "id",
+                in: "path",
+                required: true,
+                description: "Summit ID or slug",
+                schema: new OA\Schema(type: "string")
+            ),
```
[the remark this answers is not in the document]
mulldug: This comment was based on a previous version which erroneously included changes from another branch.  Those changes are no longer present in this branch.

app/ModelSerializers/Summit/SummitSerializer.php
```diff
@@ -320,24 +320,26 @@ public function serialize($expand = null, array $fields = [], array $relations =
...
-                    )->serialize(AbstractSerializer::filterExpandByPrefix($expand, 'payment_profiles'));
-
+                try {
+                    $profile = $build_default_payment_gateway_profile_strategy->build(IPaymentConstants::ApplicationTypeRegistration);
+                    $serializer = SerializerRegistry::getInstance()->getSerializer($profile, $this->getSerializerType());
+                    if (!is_null($serializer))
+                        $values['payment_profiles'][] = $serializer->serialize(AbstractSerializer::filterExpandByPrefix($expand, 'payment_profiles'));
+                } catch (\Exception $ex) {
+                    Log::warning($ex->getMessage());
+                }
```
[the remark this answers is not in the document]
mulldug: This comment was based on a previous version which erroneously included changes from another branch.  Those changes are no longer present in this branch.

app/Http/Middleware/OAuth2BearerAccessTokenRequestValidator.php
mulldug: Done.

app/Http/Middleware/OAuth2BearerAccessTokenRequestValidator.php
mulldug: Done.

PR #546, opened 2026-05-18
feat(promo-codes): Track 1 telemetry + lookup-build for DA discover

caseylocker:
ref: https://app.clickup.com/t/86ba0b8tz

## Summary

Ships **Track 1** of the Domain-Authorized Promo Codes storage-and-lookup SDS per [ftn-docsnsklz PR #15](https://github.com/fntechgit/ftn-docsnsklz/pull/15). No schema change, no Redis. Two new classes, one new trait method, a repository decouple that drops `$email` from the leaf signature and moves email filtering into the service, plus a single observability emit that becomes the Track-2 trigger instrumentation.

- New `models\summit\AllowedEmailDomainsLookup` (final readonly DTO: `exactSet` / `suffixList` / `patternsHash` / `unrestricted`).
- New `App\Services\Model\AllowedEmailDomainsLookupBuilder` (pure builder, no DB / cache / Doctrine).
- New trait method `IDomainAuthorizedPromoCode::matchesEmailDomainViaLookup(string $email, AllowedEmailDomainsLookup $lookup): bool` — O(1) exact-set + linear suffix scan. Parity-tested against legacy `matchesEmailDomain` for every pattern shape and casing.
- Repository decouple (option-(b)): `getDomainAuthorizedDiscoverableForSummit` drops `$email` and the PHP filter loop; returns raw date-windowed candidates. Aggregator `getDiscoverableByEmailForSummit` rewritten with an inline legacy filter so its public by-email contract is preserved (marked `@deprecated` for Track 2 cleanup).
- `SummitPromoCodeService::discoverPromoCodes` rewritten — calls the two leaf methods directly, builds per-code `AllowedEmailDomainsLookup` with a pattern-hash-keyed cache so sibling codes sharing pattern arrays reuse a single normalized lookup, filters via `matchesEmailDomainViaLookup`. Eliminates the ~7,000 `strtolower(trim())` calls per OCP-volume discover request (~35 ms → sub-ms).
- Track-1 observability emit: `Log::info('promo_code.discover.track1', [...])` with exactly the 4 attrs the SDS specifies: `summit_id`, `code_count`, `duration_ms`, `valid_until_date_passed`. The `valid_until_date_passed` flag is the indexed-SQL self-termination signal JP's verified ops precondition relies on.

**Companion PR:** [fntechgit/summit-admin#947](https://github.com/fntechgit/summit-admin/pull/947) (Tier 1 bulk-input frontend) — same prod push, week of 2026-05-19.

**Out of scope (Track 2, decoupled — target branch cut 2026-06-18):** `PromoCodeEmailDomainPattern` child table + migration, candidate-decomposition matcher (`EmailDomainCandidateBuilder`), batched `getTicketCountsByMemberAndPromoCodes`, trait getter/setter delegation. No Redis at any track under pass-4.

**ClickUp:** https://app.clickup.com/t/86ba0b8tz

## smarcet review pass — 2026-05-19 (commit `d7cdc3bb5`)

All 5 review findings applied in one commit on `discoverPromoCodes`. Inline replies tying each fix to file:line landed on the original review threads.

1. **Whitespace-only email bypassed the `empty()` guard** (`empty("   ") === false`, so the empty-string flowed into the email-linked repo query). Guard now reads `if (empty(trim($email))) return [];` at L1034, mirroring the legacy aggregator's normalize-first order at `DoctrineSummitRegistrationPromoCodeRepository.php:677`.
2. **Per-code lookup rebuild defeated the SDS-promised micro-opt.** Added `$lookupCache` keyed by `sha1(implode("\0", $patterns))` at L1042-1054 so sibling codes with byte-identical pattern arrays reuse a single built lookup. Pre-hash is order- and case-sensitive — acceptable for upstream-generated arrays; future T1-Sanity can surface lower-than-expected hit rate if a refinement to normalize patterns before hashing is needed.
3. **`new \DateTime()` not explicitly UTC.** L1083 now uses `new \DateTime('now', new \DateTimeZone('UTC'))` matching the repository at L699. Note: `config/app.php:70` sets `'timezone' => 'UTC'` and PHP `\DateTime` comparison normalizes via Unix timestamp regardless of TZ, so the runtime behavior was already correct — change is defensive consistency.
4. **`valid_until_date_passed` two semantic bugs.** L1084-1094 now filters `$candidates` to a `$finiteCandidates` set (non-null `getValidUntilDate()`); empty-finite case defaults to `false` instead of `true` (resolves Bug A — zero DA codes ≠ all expired); open-ended codes no longer collapse the accumulator (resolves Bug B). The reduce predicate simplifies to `$acc && $c->getValidUntilDate() < $now` since the filter pre-enforces the invariants.
5. **Test coverage gap on negative match path.** All 4 prior discovery tests mocked `matchesEmailDomainViaLookup` → `true`. Added `testDomainNonMatchingDACodeIsExcluded` using the real `AllowedEmailDomainsLookupBuilder` (already wired into `buildService()` at L80) plus Mockery `makePartial()` so the real `matchesEmailDomainViaLookup` trait method fires. Also added `testDiscoverReturnsEmptyForWhitespaceOnlyEmail` with `shouldNotReceive` on both repo methods to pin the fix-1 guard.

## Test plan

- [x] CI: `vendor/bin/phpunit` green for the 3 hot-path suites — **30/30** as of `d7cdc3bb5`:
  - `tests/Unit/Services/AllowedEmailDomainsLookupBuilderTest.php` (12/12)
  - `tests/Unit/Services/MatchesEmailDomainViaLookupTest.php` (11/11)
  - `tests/Unit/Services/SummitPromoCodeServiceDiscoveryTest.php` (**7/7** — was 5; +2 from smarcet review pass)
- [x] DoD verification — leaf method body contains zero `matchesEmailDomain(` calls (aggregator exempt):
  ```bash
  awk '/^    public function getDomainAuthorizedDiscoverableForSummit\(/,/^    }/' \
    app/Repositories/Summit/DoctrineSummitRegistrationPromoCodeRepository.php \
    | grep -c 'matchesEmailDomain('
  # must print: 0
  ```
- [ ] Behavior parity — existing discover-endpoint integration tests pass unmodified
- [ ] Local smoke against seeded summit:
  ```bash
  php artisan tinker
  $svc = app(\App\Services\Model\ISummitPromoCodeService::class);
  $result = $svc->discoverPromoCodes($summit, $member);
  # tail laravel.log for the promo_code.discover.track1 emit
  ```

## Track 1 staging sanity measurement (REQUIRED before prod merge)

Per SDS Task T1-Sanity — paste results here before approving.

1. Seed staging with one OCP-realistic DA code (`valid_until_date` in future, ~1,400 patterns, ~100% `@domain` per JP's OCP data).
2. Hit `GET /api/v1/summits/{id}/promo-codes/all/discover` as a member whose email matches a pattern; tail laravel.log for `promo_code.discover.track1`.
3. Confirm `duration_ms ≈ 35 ms`, `valid_until_date_passed = false`, `code_count = 1`.
4. SQL-flip the code's `valid_until_date` to the past, re-hit the endpoint; confirm `valid_until_date_passed` flips to `true` and `code_count → 0`.
5. Paste both measurements into this PR description.

If anything is unexpected (duration wildly higher than ~35 ms, or `valid_until_date_passed` doesn't flip), pause and escalate to the Track-2 owner — the Track-1 posture assumption is broken.

## Summary by CodeRabbit

* **New Features**
  * Domain-based promo code discovery using precomputed allowed-domain lookups for faster, consistent matching.

* **Improvements**
  * More accurate promo-code filtering (merges domain-authorized and email-linked candidates, normalizes emails, skips invalid/whitespace inputs).
  * Deterministic handling of allowed-domain patterns and stable identity detection.

* **Tests**
  * Extensive unit tests covering pattern partitioning, matching parity, discovery flows, and edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenStackweb/summit-api/pull/546?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/Repositories/Summit/DoctrineSummitRegistrationPromoCodeRepository.php
```diff
@@ -669,24 +675,29 @@ public function getDiscoverableByEmailForSummit(Summit $summit, string $email):
...
-        );
+        $daMatched = [];
+        foreach ($daCandidates as $code) {
+            if ($code instanceof IDomainAuthorizedPromoCode && $code->matchesEmailDomain($email)) {
+                $daMatched[] = $code;
+            }
+        }
+
+        $emailLinked = $this->getEmailLinkedDiscoverableForSummit($summit, $normalized);
+        return array_merge($daMatched, $emailLinked);
```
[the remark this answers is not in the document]
caseylocker: Resolved in Commit 5f7c26a0

app/Services/Model/Imp/SummitPromoCodeService.php
```diff
@@ -1024,34 +1033,69 @@ public function discoverPromoCodes(Summit $summit, Member $member): array
         $email = $member->getEmail();
         if (empty($email)) return [];
```
smarcet: **[BLOCKER] Whitespace-only email bypasses guard → live DB query with `email=\'`**

`empty("   ")` returns `false` in PHP, so a member whose `getEmail()` returns whitespace passes this check. `$normalizedEmail` becomes `""` and `getEmailLinkedDiscoverableForSummit($summit, "")` executes a real DB query binding `email=''`.

The prior aggregator had an explicit guard at `DoctrineSummitRegistrationPromoCodeRepository.php:677` (`if ($normalized === '' ) return []`) that was removed without a replacement here.

**Fix:**
```php
if (empty(trim($email))) return []; ```

Also add a unit test in `SummitPromoCodeServiceDiscoveryTest` where `$member->getEmail()` returns `"   "`, asserting the result is `[]` and no repository method is called.

File: app/Services/Model/Imp/SummitPromoCodeService.php [hunk 3]
```diff
@@ -1024,34 +1033,69 @@ public function discoverPromoCodes(Summit $summit, Member $member): array
...
-            // Global exhaustion: finite code with quantity_used >= quantity_available.
-            // The repository filter uses isLive() (dates only), so exhausted codes leak through.
-            // Skip them here so discovery matches checkout's validate() behavior.
-            if (!$code->hasQuantityAvailable()) {
-                continue;
+        // 2. Build per-code lookup once; filter via O(1) set membership + small suffix scan.
+        $matched = [];
+        foreach ($candidates as $code) {
+            if (!$code instanceof IDomainAuthorizedPromoCode) continue;
+            $lookup = $this->lookup_builder->build($code->getAllowedEmailDomains());
```
smarcet: **[BLOCKER] Lookup rebuilt on every loop iteration — O(N) optimisation does not materialise**

`$this->lookup_builder->build($code->getAllowedEmailDomains())` is called once per candidate inside the `foreach`. For a summit with 50 DA codes × 1,400 patterns each, this triggers 50 full rebuilds — each running 1,400 normalizations, a dedup pass, `sort()`, and `sha1()`. The performance gain described in the PR only applies when the lookup is built once per unique pattern-set and reused.

**Fix:** add a local `patternsHash`-keyed cache before calling `build()`:
```php
$lookupCache = [];
foreach ($candidates as $code) {
    if (!$code instanceof IDomainAuthorizedPromoCode) continue;
    $patterns = $code->getAllowedEmailDomains();
    $hash     = sha1(implode("\0", $patterns)); // stable, cheap pre-hash
    $lookup   = $lookupCache[$hash] ?? ($lookupCache[$hash] = $this->lookup_builder->build($patterns));
    if ($code->matchesEmailDomainViaLookup($normalizedEmail, $lookup)) {
        $matched[] = $code;
    }
}
```
This reduces N lookups to at most K unique pattern-sets.

app/Services/Model/Imp/SummitPromoCodeService.php
```diff
@@ -1024,34 +1033,69 @@ public function discoverPromoCodes(Summit $summit, Member $member): array
...
             }
 
             $results[] = $code;
         }
 
+        // 5. Track-1 metric emit. $candidates is the pre-filter date-windowed set;
+        //    valid_until_date_passed signals indexed-SQL self-termination per the
+        //    SDS Verified Ops Precondition.
+        $duration_ms = (microtime(true) - $started) * 1000;
+        $now = new \DateTime();
```
smarcet: **[BLOCKER] `new \DateTime()` without UTC → `valid_until_date_passed` wrong on non-UTC servers**

The repository query (line 699) creates `$now` as `new \DateTime('now', new \DateTimeZone('UTC'))`. This line uses `new \DateTime()` which inherits the PHP process system timezone. On any server where the default timezone is not UTC, `$c->getValidUntilDate() < $now` compares a UTC-stored `DateTime` against wall-clock local time — producing incorrect flag values and potentially triggering wrong ops self-termination signals.

**Fix:**
```php
$now = new \DateTime('now', new \DateTimeZone('UTC'));
```

app/Services/Model/Imp/SummitPromoCodeService.php
```diff
@@ -1024,34 +1033,69 @@ public function discoverPromoCodes(Summit $summit, Member $member): array
...
+        $valid_until_date_passed = empty($candidates)
+            ? true
+            : array_reduce(
+                $candidates,
+                fn($acc, $c) => $acc
+                    && method_exists($c, 'getValidUntilDate')
+                    && $c->getValidUntilDate() !== null
+                    && $c->getValidUntilDate() < $now,
+                true
+              );
```
smarcet, line 1096: **[BLOCKER] `valid_until_date_passed` has two semantic bugs**

**Bug A — empty candidates emits `true`**
When `$candidates` is empty (no DA codes configured for the summit), the flag short-circuits to `true`. The SDS defines this flag as "all DA codes have passed their expiry window." Zero candidates means no DA codes exist at all — not that they expired. Ops tooling acting on `true` will incorrectly self-terminate DA-code processing for summits that never had any DA codes.

**Fix:** change the empty branch to `false` (or `null` if a third state is needed):
```php
$valid_until_date_passed = empty($candidates)
    ? false   // no DA codes configured — not the same as all expired
    : array_reduce(...);
```

**Bug B — open-ended codes (null `valid_until_date`) force flag to `false`**
The reduce requires `$c->getValidUntilDate() !== null`. Any open-ended code (perpetual, no expiry set) short-circuits the accumulator to `false`. A summit with one open-ended code and several fully-expired finite codes will emit `valid_until_date_passed = false`, hiding the fact that all finite codes are past their window.

**Fix:** compute the flag only over codes that have a non-null `valid_until_date`:
```php
$finiteCandidates = array_filter($candidates, fn($c) => method_exists($c, 'getValidUntilDate') && $c->getValidUntilDate() !== null);
$valid_until_date_passed = empty($finiteCandidates)
    ? false
    : array_reduce(
        $finiteCandidates,
        fn($acc, $c) => $acc && $c->getValidUntilDate() < $now,
        true
      );
```

app/Services/Model/Imp/SummitPromoCodeService.php
```diff
@@ -1024,34 +1033,69 @@ public function discoverPromoCodes(Summit $summit, Member $member): array
...
-            // The repository filter uses isLive() (dates only), so exhausted codes leak through.
-            // Skip them here so discovery matches checkout's validate() behavior.
-            if (!$code->hasQuantityAvailable()) {
-                continue;
+        // 2. Build per-code lookup once; filter via O(1) set membership + small suffix scan.
+        $matched = [];
+        foreach ($candidates as $code) {
+            if (!$code instanceof IDomainAuthorizedPromoCode) continue;
+            $lookup = $this->lookup_builder->build($code->getAllowedEmailDomains());
+            if ($code->matchesEmailDomainViaLookup($normalizedEmail, $lookup)) {
```
smarcet, line 1052: **[FOLLOW-UP] Domain-non-matching DA code exclusion not tested at service level**

All four unit tests in `SummitPromoCodeServiceDiscoveryTest` mock `matchesEmailDomainViaLookup` to return `true`. No test verifies that a DA code returned by the repository but **not** matching the member's email domain is excluded from results.

A regression in this build-and-filter path would silently expose all in-date DA codes to every member.

**Suggested test:**
```php
public function testDomainNonMatchingDACodeIsExcluded(): void
{
    // DA code with @acme.com pattern, member has @other.com email
    $daCode = $this->buildDAPromoCode(allowedDomains: [@acme.com]); $this->repository->shouldReceive('getDomainAuthorizedDiscoverableForSummit')->andReturn([$daCode]); $this->repository->shouldReceive('getEmailLinkedDiscoverableForSummit')->andReturn([]); $member = $this->buildMember(email: 'user@other.com'); $result = $this->service->discoverPromoCodes($this->summit, $member); $this->assertEmpty($result); }
```
Use real `AllowedEmailDomainsLookupBuilder` (already wired in `buildService()`) so the test exercises the actual lookup path, not a mock.

smarcet (changes requested):
@caseylocker please review

app/Services/Model/Imp/SummitPromoCodeService.php
caseylocker: Fixed in d7cdc3bb5. `SummitPromoCodeService.php:1034` now reads `if (empty(trim($email))) return [];` per your suggested form — normalize-then-test mirrors the legacy aggregator at `DoctrineSummitRegistrationPromoCodeRepository.php:677`. Coverage added in `SummitPromoCodeServiceDiscoveryTest::testDiscoverReturnsEmptyForWhitespaceOnlyEmail` — mocks `getEmail()` to return `"   \t\n  "` and uses `shouldNotReceive` on both repo methods so any guard regression fails immediately.

app/Services/Model/Imp/SummitPromoCodeService.php
caseylocker: Fixed in d7cdc3bb5. `SummitPromoCodeService.php:1042-1054` adds a `$lookupCache` keyed by `sha1(implode("\0", $patterns))` per your suggested form. Caveat worth flagging: the pre-hash is order- and case-sensitive (no normalization before hashing), so two codes whose pattern arrays differ only in case or order won't share a cache entry. Acceptable for the OCP case where sibling codes are populated from the same upstream pattern array — but if the T1-Sanity measurement shows lower-than-expected hit rate, the fix is to normalize patterns before hashing.

app/Services/Model/Imp/SummitPromoCodeService.php
caseylocker: Fixed in d7cdc3bb5. `SummitPromoCodeService.php:1083` now uses `new \DateTime('now', new \DateTimeZone('UTC'))` matching the repository at `DoctrineSummitRegistrationPromoCodeRepository.php:699`.

Note for the record: `config/app.php:70` sets `'timezone' => 'UTC'` and PHP `\DateTime` comparison normalizes via Unix timestamp regardless of TZ, so the runtime comparison result was already correct on this app. Applied as defensive consistency — matches the repo convention and future-proofs against any config drift.

app/Services/Model/Imp/SummitPromoCodeService.php
caseylocker: Fixed in d7cdc3bb5. `SummitPromoCodeService.php:1084-1094` applies your combined fix:

```php
$finiteCandidates = array_filter(
    $candidates,
    fn($c) => method_exists($c, 'getValidUntilDate') && $c->getValidUntilDate() !== null
);
$valid_until_date_passed = empty($finiteCandidates)
    ? false
    : array_reduce(
        $finiteCandidates,
        fn($acc, $c) => $acc && $c->getValidUntilDate() < $now,
        true
      );
```

Filter to finite codes pre-enforces the non-null invariant so the reduce predicate simplifies. Empty-finite case defaults to `false` (resolves Bug A — no DA codes configured ≠ all expired). Open-ended codes no longer drag the accumulator to `false` (resolves Bug B).

app/Services/Model/Imp/SummitPromoCodeService.php
caseylocker: Fixed in d7cdc3bb5. Added `SummitPromoCodeServiceDiscoveryTest::testDomainNonMatchingDACodeIsExcluded` — uses the real `AllowedEmailDomainsLookupBuilder` already wired into `buildService()` at L80 and lets the real `matchesEmailDomainViaLookup` trait method fire (via Mockery's `makePartial()`). Member email `user@other.com` against a code allowing `@acme.com` asserts the code is excluded from results. `shouldNotReceive('setRemainingQuantityPerAccount')` additionally proves the code is dropped at the email-match step, not later at the quantity loop.

Hot-path suite now 30/30 (Builder 12 + parity 11 + Discovery 7).

smarcet (approved):
LGTM

PR #547, opened 2026-05-19
fix: reduces update logs that contain no actual changes

andrestejerina97:
ref: https://app.clickup.com/t/86b9xe3fk

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audit log handling to suppress entries when entities are updated with no meaningful changes, preventing generation of audit records for effectively unchanged data.

* **Refactor**
  * Enhanced audit log equality detection to more accurately identify when field values are unchanged across various data types.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenStackweb/summit-api/pull/547?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


andrestejerina97:
@caseylocker it's ready to review

caseylocker (approved):
LGTM to land this, closes the SelectionPlan screenshot case Mark reported.

@smarcet , the following came out of a second-pass review and extends the fix to more fully cover Mark's "widespread, not just Selection Plans" hypothesis and the preferred shape ("return a null description").
It's your call since the current pr already addresses the specific clickup ticket. Just wanted to be complete.

## Follow-up plan (separate PR)

**Correctness gaps in the current implementation**
- `AbstractAuditLogFormatter.php:228-238` — `getTimestamp()` returns integer seconds, so the `format('U.u')` fallback is unreachable and same-second microsecond changes are wrongly suppressed. Collapse to a single `format('U.u') === format('U.u')` compare.
- `AbstractAuditLogFormatter.php:240-242` — scalar branch reuses `formatChangeValue()` for equality, so `null` vs `"null"`, `true` vs `"true"`, and `1` vs `"1"` all compare equal. Replace with: null pair → strict `===`; otherwise require `gettype` match plus strict `===`. No string-coercion fallback.
- `AuditLogOtlpStrategy.php:65-71` — `hasMeaningfulChanges()` isn't on `IAuditLogFormatter`, and `AuditLogFormatterFactory.php:163` can instantiate config-driven formatters that don't extend the abstract. Drop the pre-gate; switch the existing `is_null($description)` check at line 73 to `empty($description)` so null-bubble from `format()` does the work.

**Coverage gaps (Mark's "widespread" concern)**
- `EntityUpdateAuditLogFormatter`: apply `valuesAreEffectivelyEqual` before every branch (child-entity at lines 103-108, BaseEntity at 111-142, plain at 145-157), so same-instance objects and same-instant datetimes are skipped while different entity instances still log. Return `null` on empty result. This closes the OTLP fallback case AND the DB-strategy noise for `SummitEvent` / `SummitAttendeeBadge` (the only entities `AuditLoggerFactory` resolves).
- Make `buildChangeDetails(): ?string` return `null` instead of the `"properties without changes registered"` sentinel, then update each of the ~85 callers (and `BasePresentationAuditLogFormatter::formatUpdate(): ?string`) to propagate the null so each concrete `format()` returns `null` on no-op. That lets both strategies' existing `is_null`/`empty` guards
 do the suppression — matching Sebastian's directive structurally and removing the need for a strategy-level gate.

**Polish**
- Trailing whitespace at `AuditLogOtlpStrategy.php:86`, indent of `try` at line 53, blank EOF in the new OTLP test (`AuditLogOtlpStrategyNoOpSuppressionTest.php:93`).
- `DefaultEntityManyToManyCollectionDeleteAuditLogFormatter.php:73` emits `"Removed IDs: []"` when `$deletedCount === 0`. Same family of audit noise; return `null` instead. Outside the original ticket but cheap to fix together — call out as adjacent cleanup in the PR description.

**Tests to add**
- Microsecond DateTime change → logged.
- `null` vs `"null"`, `true` vs `"true"`, `false` vs `"false"`, `1` vs `"1"` → logged.
- Ignored-fields-only change_set → null description, no dispatch on both strategies.
- `EntityUpdateAuditLogFormatter`: same-instance BaseEntity → skipped; different-instance BaseEntity (same id) → logged; tz-noise DateTime → skipped; real DateTime change → logged.
- OTLP suppresses empty-string description (not just null).
- DB strategy does not call `createAuditLogEntry` when `format()` returns null.
- PrePaid + Sponsor discount-code formatters return null on no-op (they currently add "current state" context).

smarcet:
@Andres-Tejerina — review note on the approach.

**The fix at lines 65–71 should not exist.** The strategy already has the right null-check at line 73; the pre-check duplicates logic that belongs inside the formatter.

**Root cause of the pattern:**
`buildChangeDetails()` returns the string `'properties without changes registered'` when no fields survive the filter, instead of `null`. Because it never returns `null`, `format()` never returns `null` for a no-op update, so the existing guard at line 73 never fires for this case. The PR works around that by adding `hasMeaningfulChanges()` as a parallel pre-pass — but that's two methods doing equivalent work.

**Proposed fix:**

1. Change `buildChangeDetails()` return type to `?string` — return `null` instead of `NO_CHANGES_REGISTERED_MESSAGE` when no fields survive the filter.

```php
// AbstractAuditLogFormatter
protected function buildChangeDetails(array $change_set): ?string
{
    // ... same loop, unchanged ...

    if (empty($changed_fields)) {
        return null; // was: return self::NO_CHANGES_REGISTERED_MESSAGE
    }

    return count($changed_fields) . ' field(s) modified: ' . implode(' | ', $changed_fields);
}
```

2. Each concrete formatter propagates the null out of `format()`:

```php
case IAuditStrategy::EVENT_ENTITY_UPDATE:
    $details = $this->buildChangeDetails($change_set);
    if ($details === null) return null; // no meaningful scalar change — skip entry
    return sprintf('Entity (%s) updated: %s by %s', $id, $details, $this->getUserInfo());
```

3. Remove `hasMeaningfulChanges()` from `AbstractAuditLogFormatter` and the block at lines 65–71 from `AuditLogOtlpStrategy`. The guard at line 73 already handles null and skips the job dispatch — nothing new needed in the strategy.

The `valuesAreEffectivelyEqual()` logic in `formatFieldChange()` is correct and should stay — that is what suppresses unchanged scalar fields. Only the signalling path needs fixing: unchanged → null from `buildChangeDetails()` → null from `format()` → early return at line 73.

smarcet (changes requested):
@andrestejerina97 please review

smarcet:
**`AuditLogOtlpStrategy::audit` — `Log::warning` on line 67**

This warning fires on every suppressed no-op update, not just on actual errors. Now that formatters return `null` for "no meaningful changes", this path is hit on every phantom-field save — the exact scenario the ticket describes. That replaces OTEL audit noise with Laravel warning log noise.

Change to `Log::debug()` — it is expected behaviour, not an error condition.

smarcet:
**Missing test for the actual ticket scenario**

The empty change-set case is now covered, but the ticket is specifically about two `DateTime` objects with the same value but different instances — how Doctrine loads and re-saves `begin_date`/`end_date` when something else in the Selection Plan changes. The `valuesAreEffectivelyEqual()` DateTime branch has no test at all.

Add a case along these lines to any of the existing formatter tests:

```php
$dt = new DateTime('2024-06-01 10:00:00');
$dt2 = new DateTime('2024-06-01 10:00:00'); // same value, different instance
$result = $formatter->format($subject, ['begin_date' => [$dt, $dt2]]);
$this->assertNull($result);
```

smarcet (changes requested):
@andrestejerina97 please re review

smarcet (approved):
LGTM

PR #548, opened 2026-05-19
fix(sponsors): dispatch SponsorshipCreated/Removed when sponsorships are attached inline via addSponsor or updateSponsor

romanetar:
ref https://app.clickup.com/t/86ba0vf07

## Summary by CodeRabbit

* **Improvements**
  * Sponsorship lifecycle events are now emitted per-sponsorship during sponsor creation and updates, providing more granular and complete event tracking.
* **Tests**
  * Added unit tests validating event dispatch types, payloads, and ordering (including creation, update, and removal events).
* **Chore**
  * Background job payload and event-type accessors added to support event processing and testing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenStackweb/summit-api/pull/548?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/Services/Model/Imp/SummitSponsorService.php
+++ b/app/Services/Model/Imp/SummitSponsorService.php
@@ -203,7 +203,7 @@ final class SummitSponsorService
      */
     public function updateSponsor(Summit $summit, int $sponsor_id, array $payload): Sponsor
     {
-        [$sponsor, $added_sponsorships, $removed_sponsorships] = $this->tx_service->transaction(function () use ($summit, $sponsor_id, $payload) {
+        [$sponsor, $added_sponsorships, $removed_sponsorships, $updated_sponsorships] = $this->tx_service->transaction(function () use ($summit, $sponsor_id, $payload) {
             Log::debug
             (
                 sprintf
@@ -251,6 +251,7 @@ final class SummitSponsorService
 
             $added_sponsorships = [];
             $removed_sponsorships = [];
+            $updated_sponsorships = [];
 
             if(isset($payload['sponsorship_id'])) {
                 $type_id = intval($payload['sponsorship_id']);
@@ -272,7 +273,9 @@ final class SummitSponsorService
                         )
                     );
                     // update the first
-                    $summit_sponsor->getSponsorships()->first()->setType($summit_sponsorship_type);
+                    $sponsorship = $summit_sponsor->getSponsorships()->first();
+                    $sponsorship->setType($summit_sponsorship_type);
+                    $updated_sponsorships[] = $sponsorship;
                 }
                 else {
                     Log::debug
@@ -358,7 +361,7 @@ final class SummitSponsorService
              SummitSponsorCreatedEventDTO::fromSummitSponsor($sponsor)->serialize(),
                 SponsorDomainEvents::SponsorUpdated);
 
-            return [$sponsor, $added_sponsorships, $removed_sponsorships];
+            return [$sponsor, $added_sponsorships, $removed_sponsorships, $updated_sponsorships];
         });
 
         foreach ($removed_sponsorships as $sponsorship) {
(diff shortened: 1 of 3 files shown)


app/Services/Model/Imp/SummitSponsorService.php
```diff
@@ -270,6 +290,7 @@ public function updateSponsor(Summit $summit, int $sponsor_id, array $payload):
                     $sponsorship = new SummitSponsorship();
                     $sponsorship->setType($summit_sponsorship_type);
                     $summit_sponsor->addSponsorship($sponsorship);
+                    $added_sponsorships[] = $sponsorship;
```
smarcet, line 303: This correctly adds `$added_sponsorships[]` for sponsorships created via the `sponsorships` array path. However, the sibling `sponsorship_id` path a few dozen lines above (where `hasSponsorships()` is true) calls `->getSponsorships()->first()->setType($summit_sponsorship_type)` to mutate an existing sponsorship's type, but adds nothing to `$added_sponsorships` or `$removed_sponsorships`. The semantically identical operation — changing a sponsorship type — will produce `SponsorshipRemoved` + `SponsorshipCreated` via this array path, but no events at all via `sponsorship_id`. Downstream sync consumers will diverge silently depending on which API path the caller uses. Consider capturing the old sponsorship into `$removed_sponsorships` and the new one into `$added_sponsorships`, or dispatching `SponsorshipUpdated` (already defined in `SponsorDomainEvents`) for the type-change case.

app/Services/Model/Imp/SummitSponsorService.php
```diff
@@ -338,6 +361,32 @@ function (array $acc, SummitSponsorship $sp): array {
 
             return $sponsor;
```
smarcet: The new `SponsorshipRemoved`/`SponsorshipCreated` dispatches here are correctly placed after the transaction commits. However, the `SponsorUpdated` job dispatch (a few lines above, inside the `tx_service->transaction()` closure, around lines 358–360 of this file) is still queued before the DB transaction commits. This is inconsistent: if the queue worker processes `SponsorUpdated` synchronously or very quickly, it may query the DB and see stale/uncommitted state. Consider moving the `SponsorUpdated` dispatch outside the closure, alongside these new events, to make the pattern consistent and safe.

app/Services/Model/Imp/SummitSponsorService.php
```diff
@@ -187,7 +204,10 @@ public function addSponsor(Summit $summit, array $payload): Sponsor
      */
     public function updateSponsor(Summit $summit, int $sponsor_id, array $payload): Sponsor
     {
-        return $this->tx_service->transaction(function () use ($summit, $sponsor_id, $payload) {
+        $added_sponsorships = [];
```
smarcet: No tests cover the new event dispatch paths introduced in this PR — neither `addSponsor`'s two branches (`sponsorship_id` / `sponsorships` array) nor `updateSponsor`'s remove + add paths verify that `PublishSponsorServiceDomainEventsJob` is queued with the correct event type and non-zero entity IDs after the transaction commits. Given the behavior change spans both methods and multiple code branches, at minimum one test per dispatch path would prevent silent regressions when the dispatch logic is touched again.

app/Services/Model/Imp/SummitSponsorService.php
```diff
@@ -338,6 +361,32 @@ function (array $acc, SummitSponsorship $sp): array {
...
+        foreach ($removed_sponsorships as $sponsorship) {
+            Log::debug(sprintf(
+                "SummitSponsorService::updateSponsor dispatching SponsorshipRemoved for sponsorship %s type %s sponsor %s",
+                $sponsorship->getId(),
+                $sponsorship->getType()->getId(),
+                $sponsor_id
+            ));
+            PublishSponsorServiceDomainEventsJob::dispatch(
+                DeletedEventDTO::fromEntity($sponsorship)->serialize(),
+                SponsorDomainEvents::SponsorshipRemoved);
```
smarcet: If `PublishSponsorServiceDomainEventsJob::dispatch()` throws mid-loop here (e.g., queue backend unavailable), the DB transaction is already committed, but only a subset of events fires. Downstream consumers end up with an incomplete picture of the state change — some sponsorships appear created without corresponding `SponsorshipRemoved` events, or vice versa. Consider wrapping both dispatch loops in a try/catch to log partial failures explicitly, or document that partial emission is accepted behavior for this flow.

smarcet (changes requested):

smarcet (approved):
LGTM

PR #554, opened 2026-06-03
fix(selection-plan): allow emoji in SubmissionPeriodDisclaimer

romanetar:
ref https://app.clickup.com/t/86b99xjvv

## Release Notes

* **New Features**
  * Selection plan disclaimer fields now support a wider range of characters, including emoji and extended special characters, enabling richer content expression and better international language support.

* **Tests**
  * Added regression test to verify proper storage and retrieval of emoji and special characters in selection plan disclaimers through the API.


smarcet (approved):
LGTM

PR #555, opened 2026-06-10
feat(events): add PUT /events/{id}/draft endpoint to save incomplete submissions

romanetar:
ref https://app.clickup.com/t/86ba3xg57

## Summary by CodeRabbit

* **New Features**
  * Added a `PUT /api/v1/summits/{id}/events/{event_id}/draft` endpoint to update event drafts for summits.

* **Behavior Changes**
  * Draft updates now use role-based validation/authorization and return `403` for unsupported roles.
  * When saving as draft, presentations can remain incomplete without switching to completed/received states; attempts to draft-persist already-published presentations return `412`.

* **Tests & Data**
  * Added controller tests for draft/incomplete and published-presentation cases.
  * Extended endpoint registration via migration/seeder; updated test data setup/cleanup.

### DIFF
--- a/app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitEventsApiController.php
+++ b/app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitEventsApiController.php
@@ -1085,9 +1085,48 @@ final class OAuth2SummitEventsApiController extends OAuth2ProtectedController
      */
     public function updateEvent($summit_id, $event_id)
     {
-        return $this->processRequest(function() use($summit_id, $event_id){
+        return $this->_updateEvent($summit_id, $event_id);
+    }
 
-            Log::debug(sprintf("OAuth2SummitEventsApiController::updateEvent summit id %s event id %s", $summit_id, $event_id));
+     #[OA\Put(
+        path: '/api/v1/summits/{id}/events/{event_id}/draft',
+        operationId: 'updateDraftEvent',
+        summary: 'Update an existing draft event',
+        description: 'Updates an existing draft event for a specific summit.',
+        security: [['summit_events_api_oauth2' => [SummitScopes::WriteSummitData, SummitScopes::WriteEventData]]],
+        x: ['required-groups' => [IGroup::SuperAdmins, IGroup::Administrators, IGroup::SummitAdministrators, IGroup::SummitRegistrationAdmins, IGroup::TrackChairs, IGroup::TrackChairsAdmins]],
+        tags: ['Summit Events'],
+        parameters: [
+            new OA\Parameter(name: 'id', in: 'path', required: true, description: 'Summit ID or slug', schema: new OA\Schema(type: 'string')),
+            new OA\Parameter(name: 'event_id', in: 'path', required: true, description: 'Event ID', schema: new OA\Schema(type: 'integer')),
+        ],
+        requestBody: new OA\RequestBody(
+            required: true,
+            content: new OA\JsonContent(ref: '#/components/schemas/UpdateSummitEventRequest')
+        ),
+        responses: [
+            new OA\Response(response: Response::HTTP_OK, description: 'Draft event updated successfully', content: new OA\JsonContent(ref: '#/components/schemas/SummitEvent')),
+            new OA\Response(response: Response::HTTP_BAD_REQUEST, description: 'Bad Request'),
+            new OA\Response(response: Response::HTTP_UNAUTHORIZED, description: 'Unauthorized'),
+            new OA\Response(response: Response::HTTP_FORBIDDEN, description: 'Forbidden'),
+            new OA\Response(response: Response::HTTP_NOT_FOUND, description: 'Not Found'),
+            new OA\Response(response: Response::HTTP_PRECONDITION_FAILED, description: 'Validation error'),
+            new OA\Response(response: Response::HTTP_INTERNAL_SERVER_ERROR, description: 'Server Error'),
+        ]
+    )]
+    /**
+     * @param $summit_id
+     * @param $event_id
+     * @return mixed
+     */
+    public function updateDraftEvent($summit_id, $event_id)
+    {
+        return $this->_updateEvent($summit_id, $event_id, true);
+    }
+
+    private function _updateEvent($summit_id, $event_id, bool $saveAsIncomplete = false)
+    {
+        return $this->processRequest(function() use($summit_id, $event_id, $saveAsIncomplete) {
 
             $summit = SummitFinderStrategyFactory::build($this->repository, $this->resource_server_context)->find($summit_id);
             if (is_null($summit))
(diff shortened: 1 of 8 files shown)


database/migrations/config/Version20260609175051.php
```diff
@@ -0,0 +1,83 @@
...
+    private const API_NAME = 'summits';
+    private const ENDPOINT_NAME = 'update-draft-event';
+    private const ENDPOINT_ROUTE = '/api/v1/summits/{id}/events/{event_id}/draft';
+
+    private const AUTH_GROUPS = [
+        IGroup::SuperAdmins,
+        IGroup::Administrators,
+        IGroup::SummitAdministrators,
+        IGroup::SummitRegistrationAdmins,
+        IGroup::TrackChairs,
```
smarcet: @romanetar remove track chairs ands registraion admins

app/Services/Model/Imp/SummitService.php
```diff
@@ -900,7 +903,7 @@ private function saveOrUpdatePresentationData(SummitEvent $event, SummitEventTyp
             $shouldClearSpeakers = isset($data['speakers']) && count($data['speakers']) == 0;
             $speakers = $data['speakers'] ?? [];
 
-            if ($event_type->isAreSpeakersMandatory()) {
+            if ((!$saveAsIncomplete || $event->isNew()) && $event_type->isAreSpeakersMandatory()) {
```
smarcet: The validation guard and the clearing mutation are in separate blocks, so the guard here does not prevent data corruption. When `$saveAsIncomplete=true` and the event already exists (`isNew()=false`), this condition short-circuits to `false` and the mandatory-speakers check is skipped entirely. However, the unconditional block below (`if ($shouldClearSpeakers) { $event->clearSpeakers(); }`) still executes. This means an admin can call `PUT /events/{id}/draft` on a published (STATUS_RECEIVED / PHASE_COMPLETE) presentation with `"speakers": []`, clear all speakers without hitting the ValidationException, and leave the published event in an invalid state that the rest of the system assumes cannot exist. The same applies to the moderator guard a few lines down. Consider restricting the draft endpoint to events that are not yet in a complete/published state, or move the clearing mutations inside the same guard as the validation.

tests/oauth2/OAuth2SummitEventsApiTest.php
```diff
@@ -477,6 +480,46 @@ public function testUpdateEvent()
...
+            [],
+            $this->getAuthHeaders(),
+            json_encode($data)
+        );
+
+        $this->assertResponseStatus(200);
+        $content = $response->getContent();
+        $event = json_decode($content);
+        $this->assertEquals('Updated Draft Title', $event->title);
+        $this->assertNotEquals('COMPLETE', $event->progress);
```
smarcet: This assertion always passes regardless of whether the draft feature works. `Presentation::PHASE_COMPLETE = 5` (integer), so after `json_decode` the field is `int(5)`. Comparing `assertNotEquals('COMPLETE', 5)` is trivially true — the string `'COMPLETE'` is never equal to the integer `5` in PHPUnit's equality check. A broken implementation that forces every save to `PHASE_COMPLETE` would still pass this test. Change to `assertNotEquals(Presentation::PHASE_COMPLETE, $event->progress)` to actually verify the draft behavior.

app/Services/Model/Imp/SummitService.php
```diff
@@ -881,17 +881,20 @@ private function saveOrUpdateSummitGroupEventData(SummitEvent $event, SummitEven
...
-    private function saveOrUpdatePresentationData(SummitEvent $event, SummitEventType $event_type, array $data)
+    private function saveOrUpdatePresentationData(SummitEvent $event, SummitEventType $event_type, array $data, bool $saveAsIncomplete = false)
     {
         if (!$event instanceof Presentation) return;
 
-        // if we are creating the presentation from admin, then
-        // we should mark it as received and complete
-        $event->setStatus(Presentation::STATUS_RECEIVED);
-        $event->setProgress(Presentation::PHASE_COMPLETE);
+        if (!$saveAsIncomplete || $event->isNew()) {
```
smarcet, line 895: ** No guard against calling `/draft` on a published or complete event**

`$saveAsIncomplete=true` disables status/progress transitions *and* mandatory field validation (see the speaker/moderator checks below), but there is no check here that the event is actually in a draft or incomplete state. An admin can call `PUT /events/{id}/draft` on a published event and:

1. Skip speaker/moderator mandatory validation (because `!saveAsIncomplete || isNew()` → false for an existing published event).
2. Clear speakers from a live event without error.

Suggested fix: add an early guard in `saveOrUpdateEvent` or here:

```php
if ($saveAsIncomplete && $event->isPublished()) {
    throw new ValidationException('Cannot save a published event as draft.');
}
```

Alternatively, route published events through the non-draft code path regardless of the `$saveAsIncomplete` flag.

smarcet (changes requested):
@romanetar please review

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitEventsApiController.php
```diff
@@ -1085,9 +1085,48 @@ public function addEvent($summit_id)
...
+    }
 
-            Log::debug(sprintf("OAuth2SummitEventsApiController::updateEvent summit id %s event id %s", $summit_id, $event_id));
+     #[OA\Put(
+        path: '/api/v1/summits/{id}/events/{event_id}/draft',
+        operationId: 'updateDraftEvent',
+        summary: 'Update an existing draft event',
+        description: 'Updates an existing draft event for a specific summit.',
+        security: [['summit_events_api_oauth2' => [SummitScopes::WriteSummitData, SummitScopes::WriteEventData]]],
+        x: ['required-groups' => [IGroup::SuperAdmins, IGroup::Administrators, IGroup::SummitAdministrators, IGroup::SummitRegistrationAdmins, IGroup::TrackChairs, IGroup::TrackChairsAdmins]],
```
smarcet: Two groups listed here do not match the migration and seeder, and one conflicts with the intended design:

- `IGroup::TrackChairs` — this endpoint is admin-only; TrackChairs are intentionally excluded from the migration/seeder authz groups. Listing them here will mislead API consumers and any auto-generated docs/SDK that reads `required-groups`.
- `IGroup::SummitRegistrationAdmins` — also absent from `Version20260609175051::AUTH_GROUPS` and `ApiEndpointsSeeder::authz_groups`. If they should not have access, remove this entry; if they should, add them to both the migration and seeder.

The annotation should mirror the actual gateway-enforced groups: `[SuperAdmins, Administrators, SummitAdministrators, TrackChairsAdmins]`.

smarcet: `IGroup::SummitRegistrationAdmins` should be removed from this annotation. It is absent from both `Version20260609175051::AUTH_GROUPS` and `ApiEndpointsSeeder::authz_groups`, so the gateway will reject any SummitRegistrationAdmin with a 403 regardless of what the spec says. This looks like a copy-paste from the `updateEvent` annotation — draft saving is admin-only and SummitRegistrationAdmins are not part of that scope.

smarcet (approved):
LGTM

PR #558, opened 2026-06-16
feat: add-on types crud

romanetar:
ref https://app.clickup.com/t/86b9rdth6

## Release Notes

* **New Features**
  * Added full CRUD for sponsorship add-on types at `/api/v1/summits/all/add-on-types` (OAuth2-protected: list, create, retrieve by id, update, delete).
* **Refactor**
  * Sponsorship add-ons now represent the add-on type via numeric `type_id` (with optional `type` expansion); add-on listings return only `id` and `name`, with updated filter/sort behavior.
* **Bug Fixes**
  * Reject requests that specify both `type_id` and `type` together.
  * Enforce add-on type name uniqueness; block deleting a type when it’s referenced.
* **Tests**
  * Added and updated API/controller test coverage for the new endpoints and validation/filter behavior.

### DIFF
--- a/app/Http/Controllers/Apis/Protected/Summit/Factories/SummitSponsorshipAddOnTypeValidationRulesFactory.php
+++ b/app/Http/Controllers/Apis/Protected/Summit/Factories/SummitSponsorshipAddOnTypeValidationRulesFactory.php
@@ -30,7 +30,7 @@ final class SummitSponsorshipAddOnTypeValidationRulesFactory extends AbstractVal
     public static function buildForUpdate(array $payload = []): array
     {
         return [
-            'name' => 'sometimes|string|max:255',
+            'name' => 'sometimes|required|string|max:255',
         ];
     }
 }
--- a/app/Http/Controllers/Apis/Protected/Summit/Factories/SummitSponsorshipAddOnTypeValidationRulesFactory.php
+++ b/app/Http/Controllers/Apis/Protected/Summit/Factories/SummitSponsorshipAddOnTypeValidationRulesFactory.php
@@ -1,6 +1,6 @@
 <?php namespace App\Http\Controllers;
-/**
- * Copyright 2025 OpenStack Foundation
+/*
+ * Copyright 2026 OpenStack Foundation
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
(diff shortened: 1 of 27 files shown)


app/ModelSerializers/Summit/SummitSponsorshipAddOnSerializer.php
```diff
@@ -23,12 +23,18 @@
 final class SummitSponsorshipAddOnSerializer extends SilverStripeSerializer
 {
     protected static $array_mappings = [
-        'Name' => 'name:json_string',
-        'Type' => 'type:json_string',
+        'Name'         => 'name:json_string',
+        'TypeId'       => 'type_id:json_int',
```
smarcet: The `type` string field (e.g. `"Booth"`) was present in the previous response and is now gone — only `type_id` (integer) is returned by default. Any existing consumer reading `add_on.type` will silently receive `undefined` with no deprecation window. Consider emitting a `type_name` field alongside `type_id` for backward compatibility, or defaulting the expand to include the type object until consumers are migrated.

app/Repositories/Summit/DoctrineSummitSponsorshipAddOnRepository.php
```diff
@@ -65,7 +67,6 @@ protected function getOrderMappings(): array
         return [
             'id'   => 'e.id',
             'name' => 'e.name',
-            'type' => 'e.type',
         ];
```
smarcet, line 71: The previous `'type' => 'e.type'` sort mapping was dropped and not replaced. The `at` alias is already available from the LEFT JOIN added in `applyExtraJoins()`, so adding `'type' => 'at.name'` here would restore sort-by-type-name for any consumer that was passing `order=type`.

database/migrations/model/Version20260615000001.php
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        // Seed the four types that correspond to the old string constants.
+        $this->addSql(<<<SQL
+            INSERT INTO `SummitSponsorshipAddOnType` (`Created`, `LastEdited`, `Name`) VALUES
+                (NOW(), NOW(), 'Booth'),
+                (NOW(), NOW(), 'Meeting_Room'),
+                (NOW(), NOW(), 'Schedule_Spot'),
+                (NOW(), NOW(), 'Signage_Spot')
+        SQL);
```
smarcet, line 62: This INSERT will throw a duplicate-key error if any of these four names already exist in `SummitSponsorshipAddOnType` (the table has a `UNIQUE` constraint on `Name`). The scenario is realistic after a manual rollback and re-apply, or in a test environment with pre-seeded data. Switching to `INSERT IGNORE INTO` or adding `ON DUPLICATE KEY UPDATE LastEdited = NOW()` would make this step safe to re-run.

app/Http/Controllers/Apis/Protected/Summit/Factories/SummitSponsorshipAddOnTypeValidationRulesFactory.php
```diff
@@ -0,0 +1,36 @@
...
+        return [
+            'name' => 'required|string|max:255',
+        ];
+    }
+
+    public static function buildForUpdate(array $payload = []): array
+    {
+        return [
+            'name' => 'sometimes|string|max:255',
+        ];
```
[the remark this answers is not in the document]
smarcet: @romanetar review and add regression test

smarcet (changes requested):
@romanetar please review

smarcet (approved):
LGTM

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitSponsorshipAddOnTypesApiController.php
```diff
@@ -0,0 +1,301 @@
...
+        ]]],
+        tags: ["Sponsorship Add-On Types"],
+        parameters: [
+            new OA\Parameter(name: "page", description: "Page number", in: "query", required: false, schema: new OA\Schema(type: "integer", default: 1)),
+            new OA\Parameter(name: "per_page", description: "Items per page", in: "query", required: false, schema: new OA\Schema(type: "integer", default: 10)),
+            new OA\Parameter(name: "filter", description: "Filter query (name==value, name=@value)", in: "query", required: false, schema: new OA\Schema(type: "string")),
+            new OA\Parameter(name: "order", description: "Order by (+id, -name)", in: "query", required: false, schema: new OA\Schema(type: "string")),
+        ],
+        responses: [
+            new OA\Response(response: Response::HTTP_OK, description: "OK"),
```
smarcet: @romanetar this is not correct 
https://openstackweb.github.io/summit-api/openapi/pr-558/ u need t o return a paginated response and define  a proper schema for the data 
id 
created
updated 
type
check https://github.com/OpenStackweb/summit-api/blob/2dc8cf0e7cd152dd672dfe4d3cf166b4f2cc73b0/app/Swagger/Countries.php and define the proper one for add on types

smarcet (changes requested):
@romanetar please review

smarcet (approved):
LGTM

PR #562, opened 2026-06-22
feat(speakers): add has_pending_presentations filter for speakers and submitters

romanetar:
ref https://app.clickup.com/t/86badvupk

## Summary by CodeRabbit

* **New Features**
  * Added filtering capability to filter speakers and submitters by pending presentation status across all relevant endpoints, including speakers list, submitters list, CSV exports, activity counts, and bulk email operations.

* **Tests**
  * Added test coverage for pending presentation filtering functionality for both speakers and submitters.


app/Repositories/Summit/DoctrineSpeakerRepository.php
```diff
@@ -252,6 +252,107 @@ protected function getFilterMappings()
                               WHERE 
                               __p10_2.summit = :summit AND
                               LOWER(__cb10_2.email) :operator LOWER(:value) )"),
+            'has_pending_presentations' =>
```
smarcet, line 255: @romanetar The `has_pending_presentations` definition doesn't match the ticket's stated goal. ClickUp 86badvupk is titled "Add Incomplete filter to speaker grid" and the goal is "send reminder emails to people who have **not finished their submission**", but this mapping puts no condition on `Presentation.progress` or `Presentation.status` — it matches any unpublished presentation with no team-list selection row. A speaker whose submission is `PHASE_COMPLETE`/`STATUS_RECEIVED` and simply awaiting selection matches `pending==true` (the new tests even create exactly that case and expect a match). During an open CFP, before any track-chair selection exists, this matches essentially every speaker/submitter, so a reminder email targeted with this filter would go to people who already finished their submission.

If "incomplete submission" is the intent, the condition should be based on submission state, e.g. `__p41.status IS NULL OR __p41.progress != <Presentation::PHASE_COMPLETE>`, instead of (or in addition to) the selection-list state. If the selection-based definition was agreed somewhere outside the ticket, can you point to it? Nothing in the ticket description, its comments, or summit-admin PR #989 redefines it. (Same applies to the mapping in `DoctrineMemberRepository.php`.)

tests/oauth2/OAuth2SummitSpeakersApiTest.php
```diff
@@ -614,6 +614,62 @@ public function testGetCurrentSummitSpeakersWithRejectedPresentations()
...
+            [],
+            [],
+            $headers
+        );
+
+        $content = $response->getContent();
+        $this->assertResponseStatus(200);
+        $speakers = json_decode($content);
+        $this->assertTrue(!is_null($speakers));
+        $this->assertTrue(count($speakers->data) > 0);
```
smarcet, line 670: @romanetar These three new tests pass even if the filter is a complete no-op. The two list tests assert only `count($data) > 0`, which also holds when the filter is silently ignored — the base query in `getSpeakersBySummit` already matches every speaker with summit activity. The count test's `baseline + 1` looks stronger, but `getUniqueActivitiesCountBySummit` applies the filter only to the inner speaker-ID subquery: with the filter ignored, the count becomes "all activities", and adding one presentation still yields exactly baseline + 1. So removing the mapping entirely, or breaking its `NOT EXISTS` clause, fails none of the new tests. (Same issue in `testGetCurrentSummitSpeakersActivitiesCountWithPendingPresentations` and in `OAuth2SummitSubmittersApiTest::testGetCurrentSummitSubmittersWithPendingPresentations`.)

Suggested fix: in each test also create a non-pending control (a published presentation, or one with a `SummitSelectedPresentation` row on the Group/Session list) and assert its speaker/submitter is **excluded** from the `pending==true` results — or assert the exact set of returned IDs.

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitSpeakersApiController.php
```diff
@@ -251,7 +251,7 @@ public function __construct
...
                 name: 'filter',
-                description: 'Filter by id, not_id, first_name, last_name, email, full_name, member_id, member_user_external_id, has_accepted_presentations, has_alternate_presentations, has_rejected_presentations, presentations_track_id, presentations_track_group_id, presentations_selection_plan_id, presentations_type_id, presentations_title, presentations_abstract, presentations_submitter_full_name, presentations_submitter_email, has_media_upload_with_type, has_not_media_upload_with_type. Operands supported: == (equal), @@ (contains), =@ (starts with).',
+                description: 'Filter by id, not_id, first_name, last_name, email, full_name, member_id, member_user_external_id, has_pending_presentations, has_accepted_presentations, has_alternate_presentations, has_rejected_presentations, presentations_track_id, presentations_track_group_id, presentations_selection_plan_id, presentations_type_id, presentations_title, presentations_abstract, presentations_submitter_full_name, presentations_submitter_email, has_media_upload_with_type, has_not_media_upload_with_type. Operands supported: == (equal), @@ (contains), =@ (starts with).',
```
smarcet, line 254: @romanetar The earlier CodeRabbit thread on this line was resolved as "Addressed in commit 1dc2b14", but the description it targeted — the protected `getSpeakers` filter parameter at line 181 — is unchanged: it still reads `'Filter by id, first_name, last_name, email, full_name, member_id, has_accepted_presentations, etc.'` while its three sibling descriptions (lines 254, 416, 536) were updated. Suggested fix: add `has_pending_presentations` there too, ideally the full enumerated list matching line 254 so the generated docs advertise the same filters on both routes.

smarcet (changes requested):
@romanetar please review

app/Repositories/Summit/DoctrineSpeakerRepository.php
romanetar: Fixed in de860aedf. Added the missing check on submission state to both mappings: the filter now also requires `__pX.progress != Presentation::PHASE_COMPLETE OR __pX.status IS NULL OR __pX.status != Presentation::STATUS_RECEIVED`, i.e. the negation of `Presentation::isSubmitted()`, combined with AND alongside the existing published/selection-list conditions in both `DoctrineSpeakerRepository` and `DoctrineMemberRepository`. Kept the selection-list condition as-is (combined, not replaced) rather than guessing at a redefinition that isn't documented anywhere — happy to revisit if the selection-based semantics were intentional and should be dropped instead.

tests/oauth2/OAuth2SummitSpeakersApiTest.php
romanetar: Fixed in b0df665e6. Each of the three tests now also seeds a negative control (a published presentation for a different speaker/submitter) and asserts on the exact set of returned/counted IDs instead of `count > 0` / `baseline + 1`. Confirmed locally that reverting the has_pending_presentations fix (de860aedf) makes the dedicated regression tests fail, and with the fix in place the full suites (`OAuth2SummitSpeakersApiTest`, `OAuth2SummitSubmittersApiTest`) pass.

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitSpeakersApiController.php
romanetar: Fixed in 2324cb92f. Synced the protected getSpeakers filter description with the full enumerated list already used on the public/submitters endpoints (including has_pending_presentations and has_published_presentations).

tests/oauth2/OAuth2SummitSpeakersApiTest.php
```diff
@@ -733,6 +733,148 @@ public function testGetCurrentSummitSpeakersWithRejectedPresentations()
         $this->assertTrue(!is_null($speakers));
     }
 
+    public function testGetCurrentSummitSpeakersWithPendingPresentations()
```
smarcet: @romanetar The new tests only exercise `has_pending_presentations==true` through the speaker-role branch on the list endpoints; the `false` case, the moderator branch, the submitters count endpoint and both `send` endpoints ship with no coverage.

Why it matters: the speaker `false` case in `DoctrineSpeakerRepository` is a hand-assembled `NOT EXISTS (...) AND NOT EXISTS (...)` string that no test ever executes, so an `OR`/`AND` slip or an unbalanced paren there leaves the whole suite green while the "No Pending Submissions" option summit-admin PR 989 adds (`pendingSubmissionsDDL`) returns wrong rows or a 500. Same for the moderator `EXISTS` (`__p42`): dropping or mis-joining it would go unnoticed. And the ticket's stated goal is the reminder email through `PUT .../speakers/all/send` and `.../submitters/all/send`, yet nothing asserts the chunk handed to `ProcessSpeakersEmailRequestJob` / `ProcessSubmittersEmailRequestJob` is narrowed by this filter.

All four have a precedent in these files to copy from:

- `false` branch: `testGetSubmittersWithSubmittedMediaUploadsWithType` (`OAuth2SummitSubmittersApiTest.php:336`) already queries `has_alternate_presentations==false`. Reuse the seeding of this test, query `has_pending_presentations==false`, and assert the complete/received speaker is returned and the incomplete one is not. Twin test in the submitters suite.
- Moderator branch: `testGetCurrentSummitSpeakersActivitiesCount` (`:2633`) seeds a moderated presentation via `setModerator`. Seed an incomplete presentation whose only link to the speaker is `setModerator($speaker)`, assert it is returned for `==true` and excluded for `==false`.
- `send`: `testSendSpeakersBulkEmailFilteredByMemberUserExternalId` (`:1040`) is the template: `Queue::fake()`, call `send` with `has_pending_presentations==true`, then `Queue::assertPushed(ProcessSpeakersEmailRequestJob::class, fn($job) => chunk === [pending speaker id])` with a complete/received control speaker that must not be in the chunk. Twin test in `OAuth2SummitSubmittersApiTest` next to `testSendSpeakersBulkEmail` (`:299`).
- Submitters count: mirror `testGetCurrentSummitSpeakersActivitiesCountWithPendingPresentations` (`:2774`) against `OAuth2SummitSubmittersApiController@getSubmittersActivitiesCount`, with the published control on `defaultMember2`.

smarcet (changes requested):
@romanetar please review

tests/oauth2/OAuth2SummitSpeakersApiTest.php
romanetar: Added the four gaps you flagged, all passing against a real DB run:

- `has_pending_presentations==false` for speakers (`testGetCurrentSummitSpeakersWithPendingPresentationsFalse`) and submitters (`testGetCurrentSummitSubmittersWithPendingPresentationsFalse`) - complete/received control returned, unfinished submission excluded.
- Moderator-only branch (`__p42`): `testGetCurrentSummitSpeakersWithPendingPresentationsModeratorOnly` seeds a presentation linked only via `setModerator()`, asserted for both `==true` and `==false`.
- `send` endpoints: `testSendSpeakersBulkEmailFilteredByHasPendingPresentations` asserts the chunk handed to `ProcessSpeakersEmailRequestJob` is exactly the pending speaker, not the complete/received control. For submitters, `SubmitterService::triggerSendEmails` forwards the raw filter to `ProcessSubmittersEmailRequestJob` instead of pre-resolving it to ids (unlike the speaker flow), so `testSendSubmittersBulkEmailFilteredByHasPendingPresentations` asserts the filter reaches the job unmodified via reflection on its private `filter` property - noted in the test's docblock.
- Submitters activities count: `testGetCurrentSummitSubmittersActivitiesCountWithPendingPresentations` mirrors the speakers one, baseline via direct repository call, `defaultMember2` as the published control.

Pushed as 58f9ac6b.

smarcet (approved):
LGTM

PR #564, opened 2026-06-25
fix(speakers): honour payload bio on createMySpeaker instead of always overriding with member bio

romanetar:
ref https://app.clickup.com/t/9014802374/86badw3a0

## Summary by CodeRabbit

* **Bug Fixes**
  * Speaker profile creation now respects submitted profile details over existing member data.
  * If no bio is provided, the speaker profile falls back to the member’s bio as expected.

* **Tests**
  * Added regression coverage for speaker creation to verify bio priority and fallback behavior.


smarcet (approved):
LGTM

PR #567, opened 2026-07-02
feature (registration): add order extra question answers to ticket CSV import

JpMaxMan:
ref: https://app.clickup.com/t/9014802374/86baraz3d
- After the row's ticket is resolved/created, answers are upserted for the row's attendee (falls back to the ticket owner for rows keyed only by ticket number). Unknown question names, order-usage questions, questions not allowed for the attendee, unknown list values and empty values are logged-and-skipped — they never fail the row.
- Persistence reuses the existing DTO path (`SummitAttendee::hadCompletedExtraQuestions`, same method `SummitAttendeeFactory::populate` delegates to). `populate()` itself is not called with a minimal payload because it has member/email side effects (`clearMember()` when no member passed) that are wrong for a partial update.
- Since that path clears-and-rebuilds the full answer set, former answers not overridden by the CSV are carried into the DTO (best-effort: answers whose question was deleted or is no longer allowed are dropped by the shared path — identical to the admin PUT-attendee flow).
- List-type questions: cells accept value name, label or raw id, `|`-separated for CheckBoxList (mirrors the `attendee_tags` convention); stored as comma-separated value ids as the persistence path and consumers (badge-print-app) expect. Multiple values on single-select types are rejected with a warning.
- `GET /summits/{id}/tickets/csv/template` emits one `extra_question:{name}` column per Ticket/Both-usage order extra question. OpenAPI descriptions and docblocks updated.

## Gating decision
Import **respects** the `hasAllowedExtraQuestions` gate from `AttendeeService::updateAttendee`. The persistence path clears-and-rebuilds all answers, so an attendee with zero allowed questions would have every existing answer wiped with nothing rebuilt — the gate is data protection, not permissions. Bypassing it gains nothing (no CSV answer could apply anyway) and risks destroying historical answers.

The per-question answer-change lock (`canChangeAnswerValue`) is also respected: the import job runs without an authenticated admin, so on summits with `allow_update_attendee_extra_questions` off, import fills blank answers but skips (warns on) changes to existing non-empty ones — matching what the platform enforces for the same actor context, pre-checked deterministically so a locked answer skips one column instead of rolling back the whole row.

## Tests
`tests/SummitOrderServiceTest.php`: answer set on newly created attendee; answer updated on existing attendee; unknown question skipped without failing the row; empty value ignored (existing answer preserved); list question stores value ids; badge feature columns still cleared + re-set alongside extra-question columns.

## Summary by CodeRabbit

* **New Features**
  * CSV ticket template downloads and imports now support `extra_question:{question name}` columns for ticket-scoped extra questions, including list/checkbox-list inputs stored as value IDs.
* **Bug Fixes**
  * Extra-question imports ignore unknown/empty values, skip order-scoped questions, and preserve stored answers when updates aren’t allowed.
  * Ensures badge-feature gating stays correct when badge grants and extra-question answers are in the same CSV row; clears cached allowed badge-feature results when needed.
* **Tests**
  * Expanded coverage for create/update, locked answers, list normalization, missing/empty values, order-scoped skipping, and badge/extra-question same-row scenarios.

### DIFF
--- a/app/Services/Model/ISummitOrderService.php
+++ b/app/Services/Model/ISummitOrderService.php
@@ -25,6 +25,9 @@ use Illuminate\Http\UploadedFile;
  */
 interface ISummitOrderService extends IProcessPaymentService
 {
+    // ticket data import csv column naming convention for order extra question answers
+    const ExtraQuestionColumnPrefix = 'extra_question:';
+
     /**
      * @param Member|null $owner
      * @param Summit $summit
--- a/app/Services/Model/Imp/SummitOrderService.php
+++ b/app/Services/Model/Imp/SummitOrderService.php
@@ -26,6 +26,8 @@ use App\Jobs\ProcessPaymentGatewayRefundJob;
 use App\Jobs\ProcessTicketDataImport;
 use App\Jobs\SendAttendeeInvitationEmail;
 use App\Jobs\Utils\JobDispatcher;
+use App\Models\Foundation\ExtraQuestions\ExtraQuestionType;
+use App\Models\Foundation\ExtraQuestions\ExtraQuestionTypeConstants;
 use App\Models\Foundation\Summit\Factories\SummitOrderFactory;
 use App\Models\Foundation\Summit\Registration\IBuildDefaultPaymentGatewayProfileStrategy;
 use App\Models\Foundation\Summit\Registration\PromoCodes\PromoCodesUtils;
@@ -4284,6 +4286,7 @@ final class SummitOrderService
             * badge_type_id (optional)
             * badge_type_name (optional)
             * one col per feature
+            * extra_question:{question name} (optional, one col per order extra question - Ticket/Both usage)
          */
 
         // validate format with col names
@@ -4583,6 +4586,11 @@ final class SummitOrderService
 
                 Log::debug(sprintf("SummitOrderService::processTicketData - got ticket %s (%s)", $ticket->getId(), $ticket->getNumber()));
 
+                // extra questions ( extra_question:{question name} columns )
+                $answers_owner = !is_null($attendee) ? $attendee : ($ticket->hasOwner() ? $ticket->getOwner() : null);
+                if (!is_null($answers_owner))
+                    $this->upsertAttendeeExtraQuestionAnswers($summit, $answers_owner, $row);
+
                 // badge data
                 if (!$badge_data_present) {
                     Log::warning("SummitOrderService::processTicketData badge data is not present stop current row processing.");
@@ -4655,6 +4663,182 @@ final class SummitOrderService
 
         Log::debug(sprintf("SummitOrderService::processTicketData deleting file %s from storage %s", $path, $this->download_strategy->getDriver()));
         $this->download_strategy->delete($path);
+    }
+
+    /**
+     * Upserts attendee extra question answers from a ticket data import row
+     * (one column per question, "extra_question:{question name}" naming convention).
+     * Unknown question names, order-scoped questions, disallowed questions and empty values are
+     * skipped, never fail the row. Mandatory-question completeness is not enforced ( admin bulk load ).
+     * Former answers not present on the row are carried over on a best effort basis: the shared
+     * persistence path ( ExtraQuestionAnswerHolder::hadCompletedExtraQuestions ) rebuilds the full
+     * answers set and drops answers whose question no longer exists or is no longer allowed for the
+     * attendee, same as the admin attendee update flow.
+     * @param Summit $summit
+     * @param SummitAttendee $attendee
+     * @param array $row
+     * @throws ValidationException
+     */
+    private function upsertAttendeeExtraQuestionAnswers(Summit $summit, SummitAttendee $attendee, array $row): void
+    {
+        $new_answers = [];
+
+        foreach ($row as $col => $value) {
+
+            if (!str_starts_with(strval($col), self::ExtraQuestionColumnPrefix)) continue;
+
+            $question_name = trim(substr($col, strlen(self::ExtraQuestionColumnPrefix)));
+            $value = trim(strval($value));
+
+            if ($value === '') {
+                Log::debug
+                (
+                    sprintf
+                    (
+                        "SummitOrderService::upsertAttendeeExtraQuestionAnswers question %s has an empty value for attendee %s, skipping it",
+                        $question_name,
+                        $attendee->getEmail()
+                    )
+                );
+                continue;
+            }
+
+            $question = $summit->getOrderExtraQuestionByName($question_name);
+            if (is_null($question)) {
+                Log::warning
+                (
+                    sprintf
+                    (
+                        "SummitOrderService::upsertAttendeeExtraQuestionAnswers question %s does not exist on summit %s, skipping it",
+                        $question_name,
+                        $summit->getId()
+                    )
+                );
+                continue;
+            }
+
+            if ($question->getUsage() === SummitOrderExtraQuestionTypeConstants::OrderQuestionUsage) {
+                Log::warning
+                (
+                    sprintf
+                    (
+                        "SummitOrderService::upsertAttendeeExtraQuestionAnswers question %s is order scoped, can not be answered per attendee, skipping it",
+                        $question_name
+                    )
+                );
+                continue;
+            }
+
+            if (!$attendee->isAllowedQuestion($question)) {
+                Log::warning
+                (
+                    sprintf
+                    (
+                        "SummitOrderService::upsertAttendeeExtraQuestionAnswers question %s is not allowed for attendee %s, skipping it",
+                        $question_name,
+                        $attendee->getEmail()
+                    )
+                );
+                continue;
+            }
+
+            if ($question->allowsValues()) {
+                // list type questions store the value id(s) ( comma separated when multi value )
+                // csv cells accept the value name/label ( "|" separated for multi value ) or the raw value id
+                $value_ids = [];
+                foreach (explode('|', $value) as $v) {
+                    $v = trim($v);
+                    if ($v === '') continue;
+                    $question_value = $question->getValueByName($v);
+                    if (is_null($question_value))
+                        $question_value = $question->getValueByLabel($v);
+                    if (is_null($question_value) && is_numeric($v))
+                        $question_value = $question->getValueById(intval($v));
+                    if (is_null($question_value)) {
+                        Log::warning
+                        (
+                            sprintf
+                            (
+                                "SummitOrderService::upsertAttendeeExtraQuestionAnswers value %s does not exist on question %s, skipping it",
+                                $v,
+                                $question_name
+                            )
+                        );
+                        continue;
+                    }
+                    $value_ids[] = $question_value->getId();
+                }
+                if (count($value_ids) === 0) continue;
+                // only CheckBoxList questions admit multiple selected values
+                if (count($value_ids) > 1 && $question->getType() !== ExtraQuestionTypeConstants::CheckBoxListQuestionType) {
+                    Log::warning
+                    (
+                        sprintf
+                        (
+                            "SummitOrderService::upsertAttendeeExtraQuestionAnswers question %s admits a single value, got %s, skipping it",
+                            $question_name,
+                            count($value_ids)
+                        )
+                    );
+                    continue;
+                }
+                $value = implode(ExtraQuestionType::QuestionChoicesCharSeparator, $value_ids);
+            }
+
+            $former_answer = $attendee->getExtraQuestionAnswerByQuestion($question);
+            $former_value = is_null($former_answer) ? '' : $former_answer->getValue();
+            if (!empty($former_value) && $former_value != $value && !$attendee->canChangeAnswerValue($question)) {
+                Log::warning
+                (
+                    sprintf
+                    (
+                        "SummitOrderService::upsertAttendeeExtraQuestionAnswers answer for question %s can not be changed by this time for attendee %s, skipping it",
+                        $question_name,
+                        $attendee->getEmail()
+                    )
+                );
+                continue;
+            }
+
+            $new_answers[$question->getId()] = $value;
+        }
+
+        if (count($new_answers) === 0) return;
+
+        if (!$attendee->hasAllowedExtraQuestions()) {
+            Log::warning
+            (
+                sprintf
+                (
+                    "SummitOrderService::upsertAttendeeExtraQuestionAnswers attendee %s does not have allowed extra questions, skipping answers",
+                    $attendee->getEmail()
+                )
+            );
+            return;
+        }
+
+        // upsert semantics: the persistence path rebuilds the full answers set, so carry over
+        // the former answers not overridden by the csv row ( best effort, see docblock )
+        $extra_questions = [];
+        foreach ($attendee->getExtraQuestionAnswers() as $former_answer) {
+            $question_id = $former_answer->getQuestionId();
+            if (!array_key_exists($question_id, $new_answers))
+                $extra_questions[] = ['question_id' => $question_id, 'answer' => $former_answer->getValue()];
+        }
+        foreach ($new_answers as $question_id => $value)
+            $extra_questions[] = ['question_id' => $question_id, 'answer' => $value];
+
+        Log::debug
+        (
+            sprintf
+            (
+                "SummitOrderService::upsertAttendeeExtraQuestionAnswers attendee %s answers %s",
+                $attendee->getEmail(),
+                json_encode($extra_questions)
+            )
+        );
+
+        $attendee->hadCompletedExtraQuestions($extra_questions);
     }
 
     /**
--- a/app/Services/Model/Imp/SummitOrderService.php
+++ b/app/Services/Model/Imp/SummitOrderService.php
@@ -4677,7 +4677,6 @@ final class SummitOrderService
      * @param Summit $summit
      * @param SummitAttendee $attendee
      * @param array $row
-     * @throws ValidationException
      */
     private function upsertAttendeeExtraQuestionAnswers(Summit $summit, SummitAttendee $attendee, array $row): void
     {
@@ -4782,11 +4781,20 @@ final class SummitOrderService
                     );
                     continue;
                 }
+                sort($value_ids);
                 $value = implode(ExtraQuestionType::QuestionChoicesCharSeparator, $value_ids);
             }
 
             $former_answer = $attendee->getExtraQuestionAnswerByQuestion($question);
             $former_value = is_null($former_answer) ? '' : $former_answer->getValue();
+
+            // list answers compare as sets: the same selection in a different order is not a change
+            if ($question->allowsValues() && !empty($former_value)) {
+                $former_value_ids = array_map('intval', explode(ExtraQuestionType::QuestionChoicesCharSeparator, $former_value));
+                sort($former_value_ids);
+                if ($former_value_ids === $value_ids) $value = $former_value;
+            }
+
             if (!empty($former_value) && $former_value != $value && !$attendee->canChangeAnswerValue($question)) {
                 Log::warning
                 (
@@ -4838,7 +4846,21 @@ final class SummitOrderService
             )
         );
 
-        $attendee->hadCompletedExtraQuestions($extra_questions);
+        try {
+            $attendee->hadCompletedExtraQuestions($extra_questions);
+        } catch (ValidationException $ex) {
+            // never abort the import on a bad extra-question payload: the remaining
+            // rows must still be processed ( the queue job does not retry the file )
+            Log::warning
+            (
+                sprintf
+                (
+                    "SummitOrderService::upsertAttendeeExtraQuestionAnswers attendee %s could not persist answers: %s",
+                    $attendee->getEmail(),
+                    $ex->getMessage()
+                )
+            );
+        }
     }
 
     /**
--- a/app/Services/Model/Imp/SummitOrderService.php
+++ b/app/Services/Model/Imp/SummitOrderService.php
@@ -4889,6 +4889,11 @@ final class SummitOrderService
             $value_ids[] = $question_value->getId();
         }
 
+        // duplicated tokens ( spreadsheet drag-fill artifacts, or the same choice spelled as
+        // name / label / raw id ) resolve to the same value id — dedupe before the single-value
+        // guard and before building the stored form
+        $value_ids = array_values(array_unique($value_ids));
+
         if (count($value_ids) === 0) return null;
 
         // only CheckBoxList questions admit multiple selected values
--- a/app/Services/Model/Imp/SummitOrderService.php
+++ b/app/Services/Model/Imp/SummitOrderService.php
@@ -75,6 +75,7 @@ use models\summit\SummitAttendeeTicketRefundRequest;
 use models\summit\SummitBadgeType;
 use models\summit\SummitBadgeViewType;
 use models\summit\SummitOrder;
+use models\summit\SummitOrderExtraQuestionType;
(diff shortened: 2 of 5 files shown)


app/Services/Model/ISummitOrderService.php
```diff
@@ -25,6 +25,9 @@
  */
 interface ISummitOrderService extends IProcessPaymentService
 {
+    // ticket data import csv column naming convention for order extra question answers
+    const ExtraQuestionColumnPrefix = 'extra_question:';
```
smarcet, line 29: @JpMaxMan

Consistency question on the CSV column naming.

`extra_question:{question name}` introduces a new column-prefix convention that has no precedent anywhere else in the codebase. `badge_features` — the existing "one column per admin-defined name" convention in this same import — uses the bare name as the header, no prefix (`SummitOrderService::processTicketData`, `$reader->hasColumn($featuresType->getName())`).

There's a second, more concrete inconsistency: the CSV export endpoint (`OAuth2SummitTicketApiController::getAllBySummitCSV`, `GET /summits/{id}/tickets/csv`) already represents these same order extra questions as columns, and it uses the bare `$question->getLabel()` — no prefix either. So export and this new import disagree on the column format for identical data: a CSV exported from the platform can't be round-tripped back through this import without manually renaming every question column to add `extra_question:`.

What's the reasoning for diverging from both of those existing patterns here? Worth capturing in the docblock or PR description either way, so it doesn't read as an arbitrary inconsistency to the next person touching this code — and worth deciding now whether import should just match what export already produces.

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -4657,6 +4665,204 @@ public function processTicketData(int $summit_id, string $filename)
...
+     * skipped, never fail the row. Mandatory-question completeness is not enforced ( admin bulk load ).
+     * Former answers not present on the row are carried over on a best effort basis: the shared
+     * persistence path ( ExtraQuestionAnswerHolder::hadCompletedExtraQuestions ) rebuilds the full
+     * answers set and drops answers whose question no longer exists or is no longer allowed for the
+     * attendee, same as the admin attendee update flow.
+     * @param Summit $summit
+     * @param SummitAttendee $attendee
+     * @param array $row
+     */
+    private function upsertAttendeeExtraQuestionAnswers(Summit $summit, SummitAttendee $attendee, array $row): void
```
smarcet, line 4688: @JpMaxMan

A note on this method (~180 lines).

It violates the Single Responsibility Principle: it concentrates unknown-question skip, order-scoped-question skip, the attendee-allowed gate, list-value resolution across three lookup strategies (name/label/id) with multi-value validation, change-lock enforcement, set-equality comparison for list answers, the carry-forward merge against former answers, and swallowing `ValidationException` from the underlying persistence call — eight independent concerns in a single function. Each rule maps to something the PR description calls out as a requirement, so the complexity isn't artificial — but concentrating all of it in one method makes each rule harder to test in isolation and easier to miss during review.

Concrete case in point: the `isAllowedQuestion()` gate (used here to skip disallowed answers, and again via `hasAllowedExtraQuestions()` to guard the whole write) depends on `SummitAttendee::getAllowedBadgeFeatures()`, which queries badge features via native SQL. Badge/feature assignment for the row happens *after* this method is called (in `processTicketData`), and nothing is synced to the DB until the whole per-row transaction closure returns — so a question restricted to a specific badge feature will always evaluate as "not allowed" when that feature is being granted in the same CSV row, silently dropping the answer.

Suggested fix:

1. In `processTicketData`, move the badge type/feature assignment block before the call to this method, then sync the attendee mid-transaction the same way this file already does elsewhere (e.g. `$this->attendee_repository->add($attendee, true)`) — so `getAllowedBadgeFeatures()` sees the row's own badge/features instead of a stale/empty result.
2. Split the method itself, e.g.:

```php
private function resolveExtraQuestionColumns(Summit $summit, SummitAttendee $attendee, array $row): array
// returns [question_id => value], applying the unknown/order-scoped/allowed/lock skips

private function resolveListQuestionValue(SummitOrderExtraQuestionType $question, string $raw_value): ?string
// resolves "|"-separated tokens to a sorted, comma-joined value-id string; null if unresolved/invalid

private function mergeWithFormerAnswers(SummitAttendee $attendee, array $new_answers): array
// carries forward existing answers not present in $new_answers, appends the rest, ready for hadCompletedExtraQuestions()
```

That would make the badge-feature-gate timing issue above testable in isolation, instead of only reachable through a full CSV import.

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -4657,6 +4665,204 @@ public function processTicketData(int $summit_id, string $filename)
...
+                    sprintf
+                    (
+                        "SummitOrderService::upsertAttendeeExtraQuestionAnswers question %s is order scoped, can not be answered per attendee, skipping it",
+                        $question_name
+                    )
+                );
+                continue;
+            }
+
+            if (!$attendee->isAllowedQuestion($question)) {
```
smarcet, line 4809: @JpMaxMan

The badge-feature-restricted gate here can silently drop a valid answer.

`isAllowedQuestion()` (and `hasAllowedExtraQuestions()`, called later in this same method) depends on `SummitAttendee::getAllowedBadgeFeatures()`, which queries badge features via native SQL against the DB. Badge/feature assignment for the row happens *after* this method returns (further down in `processTicketData`), and nothing is synced to the DB until the whole per-row transaction closure returns — so a question restricted to a specific badge feature will always evaluate as "not allowed" when that feature is being granted in the same CSV row, silently dropping the answer.

Reproduction test confirming this, run against this branch's test DB (`8cdce73`):

```php
public function testImportTicketDataBadgeFeatureRestrictedQuestionAnsweredInSameRowAsFeatureGrant()
{
    Queue::fake();

    $feature = new SummitBadgeFeatureType();
    $feature->setName('VIP');
    self::$summit->addFeatureType($feature);

    $question = $this->insertOrderExtraQuestion('VIP Perk Choice');
    $question->addAllowedBadgeFeatureType($feature);
    self::$em->persist(self::$summit);
    self::$em->flush();

    // same setup as testImportTicketDataBadgeFeaturesStillClearedAndReSet: unassigned ticket,
    // badge feature and extra question answer both set on the same CSV row for a brand new attendee
    $ticket = $this->getUnassignedTicket();

    $csv_content = <<<CSV
number,attendee_email,attendee_first_name,attendee_last_name,badge_type_name,VIP,extra_question:VIP Perk Choice
{$ticket->getNumber()},new.attendee@nowhere.com,New,Attendee,BADGE TYPE1,1,Yes
CSV;

    $service = $this->buildTicketDataImportService($csv_content);
    $service->processTicketData(self::$summit->getId(), 'tickets.csv');

    // the badge feature IS granted in this same row
    $feature_names = [];
    foreach ($ticket->getBadge()->getFeatures() as $f) {
        $feature_names[] = $f->getName();
    }
    $this->assertEquals(['VIP'], $feature_names);

    // ... so the question restricted to that same feature should be answerable too
    $attendee = App::make(ISummitAttendeeRepository::class)
        ->getBySummitAndEmail(self::$summit, 'new.attendee@nowhere.com');
    $this->assertNotNull($attendee);
    $answer = $attendee->getExtraQuestionAnswerByQuestion($question);
    $this->assertNotNull($answer, 'answer was dropped: isAllowedQuestion() evaluated the badge feature gate before the feature was flushed to DB');
    $this->assertEquals('Yes', $answer->getValue());
}
```

Result:
```
1) Tests\SummitOrderServiceTest::testImportTicketDataBadgeFeatureRestrictedQuestionAnsweredInSameRowAsFeatureGrant
answer was dropped: isAllowedQuestion() evaluated the badge feature gate before the feature was flushed to DB
Failed asserting that null is not null.

Tests: 13, Assertions: 37, Failures: 1, Skipped: 3.
```

The badge feature grant succeeds (first assertion passes), but the answer to the question restricted to that same feature is silently dropped (second assertion fails). The other 12 tests in the file are unaffected.

Suggested fix: move the badge type/feature assignment block in `processTicketData` before the call to `upsertAttendeeExtraQuestionAnswers`, and sync the attendee mid-transaction (e.g. `$this->attendee_repository->add($attendee, true)`, same pattern already used elsewhere in this file) so `getAllowedBadgeFeatures()` sees the row's own badge/features. Worth adding this test (or an equivalent one) to the suite either way — none of the existing tests cover a badge-feature-restricted question.

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -4657,6 +4665,204 @@ public function processTicketData(int $summit_id, string $filename)
...
+                    (
+                        "SummitOrderService::upsertAttendeeExtraQuestionAnswers question %s does not exist on summit %s, skipping it",
+                        $question_name,
+                        $summit->getId()
+                    )
+                );
+                continue;
+            }
+
+            if ($question->getUsage() === SummitOrderExtraQuestionTypeConstants::OrderQuestionUsage) {
```
smarcet, line 4797: @JpMaxMan

No test covers this skip path (`extra_question:` column referencing an `OrderQuestionUsage` question). It's a reachable branch — the CSV template only includes Ticket/Both usage questions, but nothing stops a hand-edited CSV, or a template captured before a question's usage was changed to `OrderQuestionUsage`, from including one of these columns. Worth a test asserting the row still processes normally and the attendee ends up with no answer for that question, rather than relying on this branch being exercised only implicitly.

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -4657,6 +4665,204 @@ public function processTicketData(int $summit_id, string $filename)
...
+            $former_value = is_null($former_answer) ? '' : $former_answer->getValue();
+
+            // list answers compare as sets: the same selection in a different order is not a change
+            if ($question->allowsValues() && !empty($former_value)) {
+                $former_value_ids = array_map('intval', explode(ExtraQuestionType::QuestionChoicesCharSeparator, $former_value));
+                sort($former_value_ids);
+                if ($former_value_ids === $value_ids) $value = $former_value;
+            }
+
+            if (!empty($former_value) && $former_value != $value && !$attendee->canChangeAnswerValue($question)) {
```
smarcet, line 4838: @JpMaxMan

No test covers this skip path either: an attendee with an existing non-empty answer, on a summit with `allow_update_attendee_extra_questions` off (the setting this import already documents respecting), where the CSV row tries to change that answer to a different value. `canChangeAnswerValue()` returns false in that case (no authenticated admin during a queued import), so the change should be skipped and the former value preserved — but nothing currently asserts that. The closest existing test (`testImportTicketDataUpdatesExtraQuestionAnswerOnExistingAttendee`) explicitly flips `setAllowUpdateAttendeeExtraQuestions(true)` to exercise the *allowed* path, so this locked branch is untested in either direction.

smarcet (changes requested):
@JpMaxMan please review

tests/SummitOrderServiceTest.php
```diff
@@ -373,4 +380,302 @@ public function testAutoAssignDifferentPrePaidTicketsUntilEmpty() {
...
+            $download_strategy,
+            App::make(ICompanyRepository::class),
+            App::make(ITagRepository::class),
+            App::make(ISummitRefundRequestRepository::class),
+            App::make(ICompanyService::class),
+            App::make(ITicketFinderStrategyFactory::class),
+            App::make(ITransactionService::class),
+            App::make(ILockManagerService::class)
+        );
+    }
```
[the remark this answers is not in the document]
smarcet: @JpMaxMan This finding is a false positive and can be dismissed.

`$this->upload_strategy->save()` is only invoked in `importTicketData()` — the HTTP upload path. These tests call `processTicketData()` directly, which never touches the upload strategy; it only uses the download strategy (`exists`/`get`/`getDriver`/`delete`, all stubbed in the helper).

Verified empirically against this branch (`8cdce73`): the six new import tests that exercise `processTicketData()` with this exact mock all pass. Mockery throws on any un-stubbed call, so if `save()` were reached on this path the suite could not go green.

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -4657,6 +4665,204 @@ public function processTicketData(int $summit_id, string $filename)
...
+
+        Log::debug
+        (
+            sprintf
+            (
+                "SummitOrderService::upsertAttendeeExtraQuestionAnswers attendee %s answers %s",
+                $attendee->getEmail(),
+                json_encode($extra_questions)
+            )
+        );
```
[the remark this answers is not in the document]
smarcet: @JpMaxMan Worth scoping this one: the debug line adds a second occurrence of an existing exposure, not a new one. `processTicketData()` already logs the entire CSV row — including every `extra_question:` cell value — at the top of the per-row transaction (pre-existing `Log::debug(... "processing row %s", json_encode($row))`). Trimming only this new line would not close the leak.

If answer values are considered sensitive enough to keep out of logs, the fix belongs at the file-wide logging level (the row dump is the primary source, and most log lines in this file also include attendee emails). Otherwise, this line is consistent with the file's existing logging practice.

app/Services/Model/ISummitOrderService.php
JpMaxMan: The prefix is deliberate: on this same import, every bare-name column is already claimed by badge features, so a question named like a feature ("VIP") would be ambiguous at parse time — export doesn't face that collision, which is why it can use bare labels. Honestly, had we anticipated additive per-name column families when badge features were conceived, both would have been namespaced (`badge feature:`, `extra question:`, …), leaving room for more.

Having considered the round-trip point further, we're going to do exactly that rather than defend the asymmetry: a **follow-up PR** (so this one isn't blocked) will adopt `badge_feature:{name}` alongside `extra_question:{name}` uniformly across the import template, the import itself (bare feature names stay accepted for existing CSVs), and the ticket CSV export — which is what actually closes the export→import round trip (import will also match questions by label, since export emits labels). The prefix rationale is in the method docblock as of da8b3d4.

app/Services/Model/Imp/SummitOrderService.php
JpMaxMan: Fixed in da8b3d4 — and your repro test is in the suite verbatim. Two layers were needed:

1. As you suggested: badge type/feature processing now runs **before** the answers upsert (restructured so badge-less rows no longer early-return past answer processing), and the upsert flushes the row's pending writes (`attendee_repository->add($attendee, true)`) before the permission gates run.
2. The flush alone only fixes the new-attendee case: `getAllowedBadgeFeatures()` result-caches for 600s under `ATTENDEES_{id}_BADGE_FEATURES` with no eviction anywhere, and `updateStatus()` on the reassign path warms it **pre-grant** earlier in the same row — so for an existing attendee the gate would still read stale features. Added `SummitAttendee::evictAllowedBadgeFeaturesCache()` and the upsert evicts after flushing. A second regression test covers the existing-attendee variant (`testImportTicketDataBadgeFeatureRestrictedQuestionSameRowForExistingAttendee`).

app/Services/Model/Imp/SummitOrderService.php
JpMaxMan: Split in da8b3d4 along exactly the seams you suggested: `resolveExtraQuestionColumns` / `resolveListQuestionValue` / `mergeWithFormerAnswers`, with `upsertAttendeeExtraQuestionAnswers` reduced to the orchestrator (column scan → flush+evict → resolve → gates → merge → persist). The badge-feature-gate timing issue is fixed and covered by tests — details on your 4731 thread.

app/Services/Model/Imp/SummitOrderService.php
JpMaxMan: Added in da8b3d4: `testImportTicketDataSkipsOrderScopedExtraQuestion` — order-usage question referenced from a hand-edited CSV; asserts the row still processes (attendee created, ticket assigned) and no answer is recorded.

app/Services/Model/Imp/SummitOrderService.php
JpMaxMan: Added in da8b3d4: `testImportTicketDataPreservesLockedAnswerWhenUpdatesDisallowed` — summit setting off, existing non-empty answer, CSV attempts a change; asserts the change is skipped and the former value preserved. The locked branch is now covered in both directions.

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -4657,6 +4665,204 @@ public function processTicketData(int $summit_id, string $filename)
...
+                // csv cells accept the value name/label ( "|" separated for multi value ) or the raw value id
+                $value_ids = [];
+                foreach (explode('|', $value) as $v) {
+                    $v = trim($v);
+                    if ($v === '') continue;
+                    $question_value = $question->getValueByName($v);
+                    if (is_null($question_value))
+                        $question_value = $question->getValueByLabel($v);
+                    if (is_null($question_value) && is_numeric($v))
+                        $question_value = $question->getValueById(intval($v));
```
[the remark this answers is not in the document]
JpMaxMan: Fixed in da8b3d4 — the raw-id path now requires `ctype_digit($v)` instead of `is_numeric`, so float/exponent forms no longer coerce into a value-id match.

tests/SummitOrderServiceTest.php
```diff
@@ -373,4 +380,304 @@ public function testAutoAssignDifferentPrePaidTicketsUntilEmpty() {
...
+        self::$em->persist(self::$summit);
+        self::$em->flush();
+
+        // SummitAttendee::canChangeAnswerValue caches per attendee id for 60 secs,
+        // drop any stale entry from a previous run ( attendee ids are reused across DB re-seeds )
+        Cache::flush();
+
+        $question = $this->insertOrderExtraQuestion('Dietary Requirements');
+        $attendee = $this->getDefaultAttendee();
+        $this->insertExtraQuestionAnswer($attendee, $question, 'Meat');
```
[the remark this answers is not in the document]
JpMaxMan: Fixed in da8b3d4 — replaced with a targeted `Cache::forget(sprintf("SummitAttendee.canChangeAnswerValue.%s", $attendee->getId()))` in both tests that depend on that cache.

app/Services/Model/Imp/SummitOrderService.php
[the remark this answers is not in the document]
JpMaxMan: Agreed with the scoping above: the per-row dump at the top of the transaction is the primary source, these values arrive via a plaintext CSV upload in the first place, and this line follows the file's existing logging practice — leaving as is. If importer log hygiene is wanted, that's a file-wide follow-up ticket rather than this line.

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -4583,80 +4587,351 @@ public function processTicketData(int $summit_id, string $filename)
...
+                    sprintf
+                    (
+                        "SummitOrderService::resolveListQuestionValue value %s does not exist on question %s, skipping it",
+                        $v,
+                        $question->getName()
+                    )
+                );
+                continue;
+            }
+            $value_ids[] = $question_value->getId();
```
smarcet, line 4889: @JpMaxMan

Executable evidence for the de-duplication gap in `resolveListQuestionValue` (also flagged by CodeRabbit on this helper) — two failing tests run against `670a9c7`, one per failure mode. Duplicated tokens (`Large|Large`, a common spreadsheet drag-fill artifact) or the same choice reached via different spellings (name / label / raw id) resolve to the same value id, and the id is appended here without de-duplication.

**Failure mode 1 — single-value question wrongly rejected:**

```php
public function testImportTicketDataDuplicatedTokenOnSingleValueQuestionIsAccepted()
{
    Queue::fake();

    $question = $this->insertOrderExtraQuestion
    (
        'Meal Preference',
        ExtraQuestionTypeConstants::RadioButtonListQuestionType,
        ['Vegan', 'Meat']
    );

    $ticket = $this->getUnassignedTicket();

    // duplicated token ( e.g. spreadsheet drag-fill artifact ): one distinct value selected,
    // must not be rejected by the single-value guard
    $csv_content = <<<CSV
number,attendee_email,attendee_first_name,attendee_last_name,extra_question:Meal Preference
{$ticket->getNumber()},new.attendee@nowhere.com,New,Attendee,Vegan|Vegan
CSV;

    $service = $this->buildTicketDataImportService($csv_content);
    $service->processTicketData(self::$summit->getId(), 'tickets.csv');

    $attendee = App::make(ISummitAttendeeRepository::class)
        ->getBySummitAndEmail(self::$summit, 'new.attendee@nowhere.com');
    $this->assertNotNull($attendee);

    $answer = $attendee->getExtraQuestionAnswerByQuestion($question);
    $this->assertNotNull($answer, 'answer was rejected: duplicated tokens resolve to the same value id and trip the single-value guard');
    $this->assertEquals(strval($question->getValueByName('Vegan')->getId()), $answer->getValue());
}
```

```
1) testImportTicketDataDuplicatedTokenOnSingleValueQuestionIsAccepted
answer was rejected: duplicated tokens resolve to the same value id and trip the single-value guard
Failed asserting that null is not null.
```

**Failure mode 2 — CheckBoxList persists the duplicated id string:**

```php
public function testImportTicketDataDuplicatedTokenOnCheckBoxListStoresUniqueValueIds()
{
    Queue::fake();

    $question = $this->insertOrderExtraQuestion
    (
        'T-Shirt Size',
        ExtraQuestionTypeConstants::CheckBoxListQuestionType,
        ['Small', 'Large']
    );

    $ticket = $this->getUnassignedTicket();

    $csv_content = <<<CSV
number,attendee_email,attendee_first_name,attendee_last_name,extra_question:T-Shirt Size
{$ticket->getNumber()},new.attendee@nowhere.com,New,Attendee,Large|Large
CSV;

    $service = $this->buildTicketDataImportService($csv_content);
    $service->processTicketData(self::$summit->getId(), 'tickets.csv');

    $attendee = App::make(ISummitAttendeeRepository::class)
        ->getBySummitAndEmail(self::$summit, 'new.attendee@nowhere.com');
    $this->assertNotNull($attendee);

    $answer = $attendee->getExtraQuestionAnswerByQuestion($question);
    $this->assertNotNull($answer);
    // stored form is unique value ids: "5,5" would render duplicated labels ( getNiceValue )
    // and resist later correction under the answer-change lock ( "5,5" != "5" reads as a change )
    $this->assertEquals(strval($question->getValueByName('Large')->getId()), $answer->getValue());
}
```

```
2) testImportTicketDataDuplicatedTokenOnCheckBoxListStoresUniqueValueIds
Failed asserting that two strings are equal.
--- Expected
+++ Actual
-'782'
+'782,782'
```

The persisted `"782,782"` also interacts badly with the answer-change lock: a later corrective import of `Large` produces `"782"`, which compares as a *change* against `"782,782"` — so on summits with `allow_update_attendee_extra_questions` off, the correction is skipped and the duplicated value sticks.

Verified fix: `$value_ids = array_values(array_unique($value_ids));` after the resolution loop and **before** the two `count()` checks (order matters — dedup must precede the single-value guard so mode 1 is accepted, and precede the stored form so mode 2 never persists). With that line both tests pass and the full `SummitOrderServiceTest` suite stays green (19/19). Worth adding both tests as regressions alongside the fix.

smarcet (changes requested):
@JpMaxMan please review
https://github.com/OpenStackweb/summit-api/pull/567#discussion_r3523804656

smarcet (approved):
LGTM

app/Services/Model/Imp/SummitOrderService.php
JpMaxMan: Fixed in d55df72 — `array_values(array_unique($value_ids))` after the resolution loop, before both count checks, exactly as verified. Both repro tests adopted as regressions verbatim. Thanks for the executable evidence — the change-lock interaction ("782,782" resisting later correction) was the non-obvious half.

PR #568, opened 2026-07-02
Fix inverted badge type null check on ticket data import

JpMaxMan:
## Summary
`SummitOrderService::processTicketData` badge creation for tickets without a badge had an inverted null check: it returned early when a badge type **was** provided, and would throw a `TypeError` on `SummitBadgeType::buildBadgeFromType(null)` when it wasn't. Net effect: import could never create a badge — only tickets that already had one could have their badge updated.

One-line fix (`!is_null` → `is_null`) plus a regression test that imports a row for a badge-less ticket with `badge_type_name` set and asserts the badge is created with the right type.

Stacked on #567 (the test reuses its import-test helpers). Retarget to `main` after that merges, or merge in order.

### DIFF
--- a/app/Services/Model/Imp/SummitOrderService.php
+++ b/app/Services/Model/Imp/SummitOrderService.php
@@ -4614,7 +4614,7 @@ final class SummitOrderService
 
                 if (!$ticket->hasBadge()) {
                     // create it
-                    if (!is_null($badge_type)) {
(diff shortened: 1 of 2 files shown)


JpMaxMan:
Folded into #567: the badge-block restructure there necessarily absorbed this one-line fix, and the regression test now lives in that branch.

PR #569, opened 2026-07-03
Namespace ticket CSV column families + shared value separator for export/import round-trip

JpMaxMan:
ref: https://app.clickup.com/t/86baraz3d
## Summary
Follow-up promised on #567's column-naming thread: rather than defend the asymmetry, both additive per-name column families are now namespaced — `badge_feature:{name}` joins `extra_question:{name}` — uniformly across the import template, the import (bare feature names remain accepted, so existing CSVs keep working), and the ticket CSV export. Import additionally matches questions by label, since export emits sanitized labels as headers.

Multi-value answers round-trip too: export cells now use the import's `|` separator (shared `ISummitOrderService::ExtraQuestionValueSeparator` const), with each stored value id resolved to its label individually — post-splitting `getNiceValue()`'s comma-joined output would corrupt labels containing commas. A CSV exported from `GET /summits/{id}/tickets/csv` now re-imports without any column renaming or cell editing.

Note: export headers and multi-value cell separators change. We have no scripted consumers of the export, so no compat shim is included. Remaining edge: a value label that itself contains `|` won't round-trip — inherent to any textual separator.

## Tests
`testImportTicketDataAcceptsPrefixedBadgeFeatureColumns` (canonical prefixed form; legacy bare form still covered by the existing badge-features test), `testImportTicketDataMatchesExtraQuestionByLabel`, and `testTicketCSVExportMultiValueAnswerRoundTripsThroughImport` — the last one exports a two-value answer and feeds the exported cell verbatim back through the import.

## Summary by CodeRabbit

* **New Features / Improvements**
  * Ticket CSV export/import now uses standardized, prefixed column headers for badge features and extra questions, with continued support for legacy unprefixed columns.
  * Multi-value extra-question answers now round-trip correctly and preserve selections.

* **Bug Fixes**
  * Extra questions can be matched by both question label and internal name during CSV import.
  * If CSV columns are ambiguous or conflicting, the import safely skips unclear entries and uses the eligible match when available.

### DIFF
--- a/app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitTicketApiController.php
+++ b/app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitTicketApiController.php
@@ -745,7 +745,7 @@ final class OAuth2SummitTicketApiController extends OAuth2ProtectedController
         path: '/api/v1/summits/{id}/tickets/csv/template',
         operationId: 'getTicketImportTemplate',
         summary: 'Get ticket import template',
-        description: 'Returns a CSV template for importing ticket data. Includes one column per summit badge feature name plus one extra_question:{question name} column per summit order extra question (Ticket/Both usage).',
+        description: 'Returns a CSV template for importing ticket data. Includes one badge_feature:{feature name} column per summit badge feature plus one extra_question:{question name} column per summit order extra question (Ticket/Both usage).',
         security: [['summit_tickets_oauth2' => [
             SummitScopes::WriteSummitData,
             SummitScopes::WriteRegistrationData,
@@ -788,8 +788,8 @@ final class OAuth2SummitTicketApiController extends OAuth2ProtectedController
              * ticket_promo_code (optional)
              * badge_type_id (optional)
              * badge_type_name (optional)
-             * badge_features (optional, one col per badge feature name)
-             * extra_question:{question name} (optional, one col per order extra question - Ticket/Both usage)
+             * badge_feature:{feature name} (optional, one col per badge feature; bare feature name accepted for legacy csvs)
+             * extra_question:{question name} (optional, one col per order extra question - Ticket/Both usage; label accepted too)
              */
 
             $summit = SummitFinderStrategyFactory::build($this->summit_repository, $this->getResourceServerContext())->find($summit_id);
@@ -814,7 +814,7 @@ final class OAuth2SummitTicketApiController extends OAuth2ProtectedController
 
             // badge features for summit
             foreach ($summit->getBadgeFeaturesTypes() as $featuresType) {
-                $row[$featuresType->getName()] = '';
+                $row[sprintf('%s%s', ISummitOrderService::BadgeFeatureColumnPrefix, $featuresType->getName())] = '';
             }
 
             // order extra questions for summit ( ticket / attendee scoped ones )
@@ -841,7 +841,7 @@ final class OAuth2SummitTicketApiController extends OAuth2ProtectedController
         path: '/api/v1/summits/{id}/tickets/csv',
         operationId: 'importTicketData',
         summary: 'Import ticket data from CSV',
-        description: 'Imports ticket data from a CSV file. Supported columns: id, number, attendee_email, attendee_first_name, attendee_last_name, attendee_tags, attendee_company, attendee_company_id, ticket_type_name, ticket_type_id, promo_code_id, promo_code, ticket_promo_code, badge_type_id, badge_type_name, one column per badge feature name (1/0) and one extra_question:{question name} column per order extra question (Ticket/Both usage; for list type questions use the value name/label, "|" separated for multi value).',
+        description: 'Imports ticket data from a CSV file. Supported columns: id, number, attendee_email, attendee_first_name, attendee_last_name, attendee_tags, attendee_company, attendee_company_id, ticket_type_name, ticket_type_id, promo_code_id, promo_code, ticket_promo_code, badge_type_id, badge_type_name, one badge_feature:{feature name} column per badge feature (1/0; bare feature name accepted for legacy CSVs) and one extra_question:{question name} column per order extra question (Ticket/Both usage; question label accepted too; for list type questions use the value name/label, "|" separated for multi value).',
         security: [['summit_tickets_oauth2' => [
             SummitScopes::WriteSummitData,
             SummitScopes::WriteRegistrationData,
--- a/app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitTicketApiController.php
+++ b/app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitTicketApiController.php
@@ -645,11 +645,12 @@ final class OAuth2SummitTicketApiController extends OAuth2ProtectedController
                     ];
 
                     foreach ($summit->getBadgeFeaturesTypes() as $featuresType) {
-                        $allowed_columns[] = $featuresType->getName();
+                        $allowed_columns[] = sprintf('%s%s', ISummitOrderService::BadgeFeatureColumnPrefix, $featuresType->getName());
                     }
 
                     foreach ($summit->getOrderExtraQuestionsByUsage(SummitOrderExtraQuestionTypeConstants::TicketQuestionUsage) as $question) {
-                        $allowed_columns[] = $question->getLabel();
+                        // must match the serializer-emitted column exactly ( prefixed, sanitized label )
+                        $allowed_columns[] = sprintf('%s%s', ISummitOrderService::ExtraQuestionColumnPrefix, html_entity_decode(strip_tags($question->getLabel())));
                     }
 
                     $columns_param = Request::input("columns", "");
--- a/app/Services/Model/Imp/SummitOrderService.php
+++ b/app/Services/Model/Imp/SummitOrderService.php
@@ -4785,28 +4785,20 @@ final class SummitOrderService
             }
 
             // match by name and by label ( the ticket csv export emits question labels as
-            // column names ); when the two lookups disagree the column is ambiguous — one
-            // question's name is another question's label — skip it rather than silently
-            // filing the answer under the wrong question
-            $question = $summit->getOrderExtraQuestionByName($question_name);
-            $question_by_label = $summit->getOrderExtraQuestionByLabel($question_name);
-            if (!is_null($question) && !is_null($question_by_label) && $question_by_label->getId() !== $question->getId()) {
-                Log::warning
-                (
-                    sprintf
-                    (
-                        "SummitOrderService::resolveExtraQuestionColumns column %s is ambiguous on summit %s ( name of question %s, label of question %s ), skipping it",
-                        $question_name,
-                        $summit->getId(),
-                        $question->getId(),
-                        $question_by_label->getId()
-                    )
-                );
-                continue;
+            // column names ); each candidate is vetted for usage / attendee eligibility
+            // before deciding ambiguity, so a question that could never take this answer
+            // ( order scoped or not allowed for the attendee ) can not shadow a legitimate
+            // match; only when two eligible questions remain — one question's name is
+            // another question's label — is the column truly ambiguous and skipped, rather
+            // than silently filing the answer under the wrong question
+            $candidates = [];
+            foreach ([$summit->getOrderExtraQuestionByName($question_name),
+                      $summit->getOrderExtraQuestionByLabel($question_name)] as $candidate) {
+                if (is_null($candidate)) continue;
+                $candidates[$candidate->getId()] = $candidate;
             }
-            if (is_null($question))
-                $question = $question_by_label;
-            if (is_null($question)) {
+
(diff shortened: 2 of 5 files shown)


app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitTicketApiController.php
romanetar, line 617: Export's columns= query-param allowlist wasn't updated to the new namespaced keys. $allowed_columns
  still pushes bare $featuresType->getName() / $question->getLabel(), never the badge_feature:/extra_question: prefixed forms the serializer now emits. Confirmed via full call-trace: the
  default (no columns=) export is unaffected since CSVExporter::export() ignores this list entirely and just uses the serialized row's actual keys — but a client explicitly requesting
  ?columns=badge_feature:VIP (the real, current column name) gets rejected with a ValidationException since that string isn't in the stale allowlist.

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -4781,6 +4785,9 @@ private function resolveExtraQuestionColumns(Summit $summit, SummitAttendee $att
             }
 
             $question = $summit->getOrderExtraQuestionByName($question_name);
```
romanetar: Extra-question name→label fallback can silently mis-file an answer to the wrong question. getOrderExtraQuestionByName() and the new getOrderExtraQuestionByLabel() fallback are independent lookups with no cross-check. If Question A's name equals Question B's label (both plausible, free-text fields with no cross-field uniqueness constraint), a CSV column extra_question:{that string} — including one produced by this PR's own export, which always emits labels as headers — resolves to Question A via the name-match and never reaches the label fallback intended for B. The answer is silently written to the wrong question; no warning is logged since a question was found. This directly undermines the PR's stated round-trip goal. (Independently confirmed by 3 of the 6 finder agents via the same mechanism.)

romanetar (changes requested):
@JpMaxMan please review comments

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitTicketApiController.php
JpMaxMan: Fixed in 970b38a — both allowlist loops now emit the namespaced forms, and the question entries use the sanitized label (`html_entity_decode(strip_tags(...))`) so they match what the serializer actually emits (the raw-label form in the old allowlist was already drifting from the emitted header for HTML-bearing labels). Confirmed your call-trace read: the default no-`columns=` path was unaffected either way.

app/Services/Model/Imp/SummitOrderService.php
JpMaxMan: Fixed in 970b38a — agreed silent mis-file is the worst outcome for an importer, and reordering the lookups would just mirror the failure mode. The fix runs both lookups side by side: when name-match and label-match disagree, the column is logged as ambiguous (naming both question ids) and skipped entirely. Regression test added (`testImportTicketDataSkipsAmbiguousExtraQuestionColumn`): question A's name == question B's label → the ambiguous column touches neither question, row still processes.

app/Services/Model/Imp/SummitOrderService.php
```diff
@@ -4780,7 +4784,28 @@ private function resolveExtraQuestionColumns(Summit $summit, SummitAttendee $att
                 continue;
             }
 
+            // match by name and by label ( the ticket csv export emits question labels as
+            // column names ); when the two lookups disagree the column is ambiguous — one
+            // question's name is another question's label — skip it rather than silently
+            // filing the answer under the wrong question
             $question = $summit->getOrderExtraQuestionByName($question_name);
```
romanetar: @JpMaxMan The ambiguity check runs before the usage/eligibility filters, causing false-positive skips. The name-vs-label collision check
  (lines 4791-4806) executes before the OrderQuestionUsage check (line 4822) and isAllowedQuestion() check (line 4834). So if a legitimately-matched Ticket-usage question's name collides
  with the label of an unrelated Order-scoped (or attendee-disallowed) question, the column is now flagged "ambiguous" and dropped entirely — even though the label-matched question would
  have been rejected by the usage/eligibility check moments later anyway and was never a real candidate. This is a regression in import completeness introduced by the fix: previously
  (name-match-only, no fallback) this exact column would have imported correctly. No test exercises this cross-usage case — only same-usage true collisions are covered.

romanetar (changes requested):
@JpMaxMan found an edge-case regression introduced by the previous fix. Please review comment.

app/Services/Model/Imp/SummitOrderService.php
JpMaxMan: Good catch — fixed in 4a69e57. The name/label lookups now build a candidate list that's vetted through the existing order-scoped and `isAllowedQuestion()` rejections first; a column is only skipped as ambiguous when two *eligible* questions survive, so an ineligible collision partner can no longer shadow a legitimate match. Added `testImportTicketDataIneligibleQuestionDoesNotMakeColumnAmbiguous` covering the cross-usage case (ticket-usage question named X vs order-scoped question labeled X → resolves to the ticket-usage question).

Worth noting for severity: order-scoped questions aren't surfaced anywhere in the registration FE today, so that half of the path is defensive — the realistic leg was the `isAllowedQuestion()` one, which is now also decided per-candidate before ambiguity.

smarcet:
@JpMaxMan Deep review pass on this PR, adversarially verified (two independent checks per finding, actively trying to disprove each one before posting).

## Label sanitization mismatch breaks the round-trip for HTML/entity-bearing labels

Export builds the header as `extra_question:` + `html_entity_decode(strip_tags($question->getLabel()))` (`SummitAttendeeTicketCSVSerializer.php:91-94`). Import resolves that header via `Summit::getOrderExtraQuestionByLabel()`, which does an **exact match against the raw stored label** (`Summit.php:4539-4545`, `Criteria::expr()->eq('label', trim($label))`) — no `strip_tags`/`html_entity_decode` anywhere on the import side.

So any question whose label contains HTML tags or entities exports a *sanitized* header that can never equal the *raw* stored label. If the question's `name` doesn't happen to match either, the column is treated as "unknown question" and the answer is silently dropped (just a debug/warning log, no error surfaced).

This isn't a hypothetical edge case:
- `SummitOrderExtraQuestionTypeService.php:327` sets `setLabel(trim($question['question']['html']))` for Eventbrite-seeded questions (and `:343` for waiver questions), both with `usage = TicketQuestionUsage` — exactly the questions this export/import round-trip targets. That job is live and wired (`IngestSummitExternalRegistrationData`, dispatched from `SummitOrderService.php:5146`).
- Even outside Eventbrite, the normal admin path runs labels through `HTMLCleaner::cleanData()` (HTMLPurifier), which by default *permits* common tags (`b`, `i`, `em`, `strong`, `a`, `p`, ...) rather than stripping everything — so an admin bolding part of a question label through the normal UI hits the same mismatch.
- `ExtraQuestionType::setLabel()` only trims whitespace — nothing sanitizes the label at save time to make this a non-issue.

Proof — this test reproduces the drop (label `Caf&eacute; Preference` → exported header `Café Preference` → re-import of that exact exported cell loses the answer):

```php
public function testTicketCSVExportHtmlEntityLabelDoesNotRoundTripThroughImport()
{
    Queue::fake();

    // label as stored e.g. by the Eventbrite ingestion path (SummitOrderExtraQuestionTypeService
    // ::seedSummitOrderExtraQuestionTypesFromEventBrite) or by an admin entering basic markup
    // through the normal API (HTMLPurifier permits <b>/<i>/<a>/... by default)
    $question = $this->insertOrderExtraQuestion
    (
        'DIET_REQ',
        ExtraQuestionTypeConstants::TextQuestionType,
        [],
        SummitOrderExtraQuestionTypeConstants::TicketQuestionUsage,
        'Caf&eacute; Preference'
    );

    $attendee = $this->getDefaultAttendee();
    $ticket = $attendee->getTickets()->first();

    // export: header is the sanitized label ( html_entity_decode(strip_tags(...)) ), not the raw one
    $values = SerializerRegistry::getInstance()
        ->getSerializer($ticket, SerializerRegistry::SerializerType_CSV)
        ->serialize(null, [], [], [
            'ticket_questions' => [$question],
            'answers_by_owner' => [$attendee->getId() => [$question->getId() => 'Vegan']],
        ]);

    $question_column = 'extra_question:Café Preference';
    $this->assertArrayHasKey($question_column, $values);
    $this->assertEquals('Vegan', $values[$question_column]);

    // re-import the exact exported header + cell, verbatim, for a new attendee
    $unassigned = $this->getUnassignedTicket();
    $csv_content = <<<CSV
number,attendee_email,attendee_first_name,attendee_last_name,{$question_column}
{$unassigned->getNumber()},new.attendee@nowhere.com,New,Attendee,Vegan
CSV;

    $service = $this->buildTicketDataImportService($csv_content);
    $service->processTicketData(self::$summit->getId(), 'tickets.csv');

    $new_attendee = App::make(ISummitAttendeeRepository::class)
        ->getBySummitAndEmail(self::$summit, 'new.attendee@nowhere.com');
    $this->assertNotNull($new_attendee);

    // getOrderExtraQuestionByLabel() matches the RAW stored label ('Caf&eacute; Preference'),
    // which never equals the sanitized exported header ('Café Preference'), so the column is
    // treated as an unknown question and the answer is silently dropped
    $answer = $new_attendee->getExtraQuestionAnswerByQuestion($question);
    $this->assertNotNull($answer, 'expected the CSV export/import round trip to preserve the answer');
    $this->assertEquals('Vegan', $answer->getValue());
}
```

Ran against the current branch head (`f9198e54d`):

```
1) Tests\SummitOrderServiceTest::testTicketCSVExportHtmlEntityLabelDoesNotRoundTripThroughImport
expected the CSV export/import round trip to preserve the answer
Failed asserting that null is not null.
```

Suggested fix: apply the same `html_entity_decode(strip_tags(...))` transform to the DB comparison (compare in PHP, not in the Criteria query), or add a dedicated sanitized-label lookup that mirrors the export transform exactly.

## Extra-question column resolution issues two DB queries per column per row

`resolveExtraQuestionColumns` (`SummitOrderService.php:4795-4796`) unconditionally calls both `getOrderExtraQuestionByName()` and `getOrderExtraQuestionByLabel()` for every `extra_question:*` column on every CSV row, with no memoization. `Summit::$order_extra_questions` is `fetch: 'EXTRA_LAZY'` (`Summit.php:589`), so each `matching()` call is a fresh SQL query rather than an in-memory filter — confirmed by tracing into `LazyCriteriaCollection::doInitialize()`. Before this PR it was one query per column per row (name-match only); now it's two, unconditionally.

I found this is a deliberate trade-off (an earlier commit on this branch short-circuited the label lookup; a later one removed the short-circuit specifically to fix the ambiguous name/label collision case), and with realistic CSVs having ~1-3 extra-question columns against a small per-summit question table in a background queued job, it's minor, not a hot-path regression. Flagging mainly as a follow-up: consider pre-loading `$summit`'s extra questions once per import job and building a local name/label index instead of two DB round-trips per column per row.

smarcet (changes requested):
@JpMaxMan please review

JpMaxMan:
@smarcet Both landed in `2198074` — thanks for the adversarially-verified pass, this is exactly the kind of review that catches what a quick look wouldn't.

**Finding 1 (HTML/entity label mismatch):** fixed by folding the by-name and by-label resolution into one per-summit lookup index (`getOrderExtraQuestionsIndex()`) keyed by name, raw label, **and** the export-sanitized label (`html_entity_decode(strip_tags(...))` — same transform the serializer applies). Added `testTicketCSVExportHtmlEntityLabelRoundTripsThroughImport`, adapted from your repro (`Caf&eacute; Preference` → exported `Café Preference` → re-import now resolves instead of silently dropping the answer).

**Finding 2 (2x EXTRA_LAZY queries per column per row):** resolved as a side effect of the same change — the index is built once per summit (cached on the service instance) instead of two `matching()` round-trips per column per row.

Re-requesting your review on this one — CI is running now.

smarcet:
@JpMaxMan please review unit tests ( CI is red )
https://github.com/OpenStackweb/summit-api/actions/runs/28919500453/job/85793392737?pr=569#step:10:816

1) Tests\SummitOrderServiceTest::testTicketCSVExportHtmlEntityLabelRoundTripsThroughImport
expected the CSV export/import round trip to preserve the answer
Failed asserting that null is not null.

/home/runner/work/summit-api/summit-api/tests/SummitOrderServiceTest.php:1088

smarcet (changes requested):
@JpMaxMan please review unit tests 
CI is red
https://github.com/OpenStackweb/summit-api/actions/runs/28919500453/job/85793392737?pr=569#step:10:816

1) Tests\SummitOrderServiceTest::testTicketCSVExportHtmlEntityLabelRoundTripsThroughImport
expected the CSV export/import round trip to preserve the answer
Failed asserting that null is not null.

/home/runner/work/summit-api/summit-api/tests/SummitOrderServiceTest.php:1088

PR #40, opened 2022-05-06
check in endpoint

romanetar:
ref https://tipit.avaza.com/project/view#!tab=task-pane&groupby=MyTaskDueDate&view=vertical&task=2916710&fileview=grid

Signed-off-by: romanetar <roman_ag@hotmail.com>

PR #571, opened 2026-07-09
feat(badges): augment validateBadge payload for attendee-networking connect resolve

smarcet:
https://app.clickup.com/t/9014802374/86bauc5hf

Extend GET /summits/{id}/badge/{badge}/validate so attendee-networking-api can resolve a scanned badge to a connection identity (SDS D26 / ftn-docsnsklz #40):

  - add ticket.owner.id (keys the member-less pre-account request, D27)
  - expand the nested ticket.owner.member {id, first_name, last_name, pic} (member.id is the D15 identity; member is absent for member-less attendees)

No auth change is needed: the auth.user middleware passes service tokens through (required-groups gate only user tokens), so a client-credentials service account holding the read scope already reaches the route.

Tests (OAuth2SummitApiTest): the member case asserts the expanded member object; the member-less case (D27) asserts the member object is absent while attendee id + email are present. InsertSummitTestData gains a member-less attendee with a badge to cover it.

### DIFF
--- a/.github/workflows/check-branch-name.yml
+++ b/.github/workflows/check-branch-name.yml
@@ -12,7 +12,7 @@ jobs:
         run: |
           BRANCH_NAME="${{ github.head_ref }}"
           echo "Checking branch name: $BRANCH_NAME"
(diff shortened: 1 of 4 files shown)


romanetar (approved):
LGTM

PR #572, opened 2026-07-15
Hotfix/attendee regenerate qr code

smarcet:
ref: https://app.clickup.com/t/9014802374/86baxuvj3

## Summary by CodeRabbit

* **New Features**
  * Badge QR codes are now regenerated when tickets are reassigned or updated.
  * Attendee badge QR codes are generated when badge details are requested through the attendee API.
  * QR code generation failures no longer interrupt ticket reassignment or updates.

* **Documentation**
  * Updated OpenTelemetry trace-viewing instructions.

* **Chores**
  * Adjusted local development ports for messaging, telemetry, and search services.

* **Tests**
  * Added coverage for QR code regeneration, persistence, and badge updates.


app/Services/Model/AttendeeService.php
```diff
@@ -554,6 +573,36 @@ public function reassignAttendeeTicket(Summit $summit, SummitAttendee $attendee,
...
+
+        if (empty($badge_ids_to_generate)) return;
+
+        foreach ($badge_ids_to_generate as $badge_id) {
+            $this->tx_service->transaction(function () use ($attendee, $badge_id) {
+                $badge = $this->badge_repository->getByIdExclusiveLock($badge_id);
+                if (!$badge instanceof SummitAttendeeBadge) return;
+                Log::debug(sprintf("AttendeeService::regenerateAttendeeBadgesQRCodes attendee %s badge %s", $attendee->getId(), $badge_id));
+                try {
+                    $badge->generateQRCode();
```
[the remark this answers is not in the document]
smarcet: Addressed 2 of the 3 points here, and explaining why the third is intentional:

**1. Skip badges with an existing QR code before/after locking** — Not applying this. Unconditional regeneration on every call is intentional (explicit product decision, not an oversight): `regenerateAttendeeBadgesQRCodes` always regenerates every active badge's QR, it does not generate-once-if-missing. `generateQRCode()` is a deterministic function of ticket/owner fields, so repeated regeneration with unchanged data produces an identical string — skipping would only save compute, not change correctness.

**2. Invoke regeneration only when `tickets.badge` is expanded** — Done in e68e373a0. `getOwnAttendee` now only calls `regenerateAttendeeBadgesQRCodes()` when `tickets.badge` is present in the `expand` param, so requests that never serialize badge data skip the exclusive-lock/regeneration work entirely.

**3. Assert the generation path is skipped, not just that values match** — Added `testGetOwnAttendeeDoesNotGenerateBadgeQRCodeWhenBadgeNotExpanded` in the same commit: asserts the badge's `qr_code` stays empty after a request without `tickets.badge` (and after `expand=tickets` alone), proving the call is skipped rather than re-deriving the same deterministic value.

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitAttendeesApiController.php
```diff
@@ -149,8 +149,8 @@ public function __construct
     #[OA\Get(
         path: '/api/v1/summits/{id}/attendees/me',
         operationId: 'getCurrentAttendee',
-        summary: 'Get current user attendee profile',
-        description: 'Returns the attendee profile for the currently authenticated user in the specified summit',
+        summary: 'Get current user attendee profile for a particular summit',
+        description: 'Returns the attendee profile for the currently authenticated user in the specified summit ( regenerates que Badge QR code of the active tickets ) ',
```
romanetar: @smarcet please fix the typo (que)

romanetar (changes requested):
@smarcet please review

app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitAttendeesApiController.php
smarcet: Fixed in f60cd0e7c — corrected the typo and updated the wording to reflect current behavior (regeneration only happens when `tickets.badge` is expanded).

romanetar (approved):
LGTM

PR #573, opened 2026-07-15
feat(events): fix and complete get-overflow-published-events endpoint

smarcet:
ref: https://app.clickup.com/t/9014802374/86baxyy0u
## Summary

Completes and fixes the `GET /api/v1/summits/{id}/events/all/published/occupancy/overflow` endpoint, which returns published events with `occupancy == OVERFLOW` for a given summit.

## Security hardening (post-review)

The endpoint's response reuses `SummitEventOverflowStreamingSerializer`, which includes `overflow_tokens` (signed Mux playback JWTs) — by design per ClickUp 86baxyy0u, which proposes this as an authenticated reconcile endpoint for cold-started clients. That ticket explicitly left the authorization scope as an open API-owner decision ("whether it should be attendee-gated or plain auth.user"). This PR originally answered that with the broad `ReadSummitData`/`ReadAllSummitData` scopes, shared by 36+ unrelated read endpoints — any client holding either one could pull live playback tokens for every overflow event in a summit in a single call.

Resolved by adding a dedicated scope instead:

- New `SummitScopes::ReadOverflowEvents` (`.../summits/events/overflow/read`).
- Endpoint now requires only this scope (OA `security` attribute, migration, and seeder updated in lockstep — scope checks are OR-based, so the broad scopes had to be replaced, not just supplemented).
- `Version20260715120000` now also inserts the `api_scopes` row for the new scope before associating it with the endpoint (reversible in `down()`).

**Deploy action required:** `Version20260715120000` only registers the scope + endpoint in summit-api's own API/scope catalog (config DB) — it does **not** grant the scope to any OAuth2 client. Any client application that needs to call this endpoint must be explicitly granted `SummitScopes::ReadOverflowEvents` (`.../summits/events/overflow/read`) on its OAuth2 client record (IDP-side) before/at deploy. Holding `ReadSummitData` or `ReadAllSummitData` alone is no longer sufficient — those clients will get `403 Insufficient Scope` on this endpoint after this ships.

## Deployment

- Added the endpoint to `ApiEndpointsSeeder` (fresh installs), gated by `ReadOverflowEvents`.
- Added `Version20260715120000` migration to register the scope + endpoint (name, route, scope) for already-deployed environments, including k8s. Verified up/down/up against the config DB.
- Generalized `APIEndpointsMigrationHelper` with `registerEndpoint(s)`/`unregisterEndpoint(s)` composite helpers so future endpoint migrations need one call per endpoint instead of repeating insert/scope/authz-group boilerplate.

## Testing

- `testGetOverflowPublishedEvents` extended with negative cases: an unpublished overflow event on the same summit, and a published overflow event on a different summit, proving the `published`/`summit_id` filters correctly exclude both.
- Full `OAuth2SummitEventsApiTest.php` suite: 56 tests, 127 assertions, 0 failures.
- `OAuth2PresentationApiTest.php` (shares the modified filter-mapping code path): 42 tests, 0 failures.

## Summary by CodeRabbit

* **New Features**
  * Added a new `GET /api/v1/summits/{id}/events/all/published/occupancy/overflow` endpoint to retrieve overflow occupancy published events.
  * Introduced `occupancy` filtering support for summit event queries.
  * Added and seeded a dedicated read permission scope for overflow events.
* **Bug Fixes**
  * Updated event streaming serialization to always reflect the latest overflow tokens.
  * Improved endpoint registration/cleanup during configuration migrations.
* **Tests**
  * Added coverage for the new overflow endpoint and verified overflow streaming serializer token freshness.


romanetar (approved):
LGTM

PR #576, opened 2026-08-03
feat(badges): accept URL-safe base64 (base64url) badge artifacts

smarcet:
ref: https://app.clickup.com/t/9014802374/86bb7wfgw
## What

Widen `App\Utils\Base64` to also accept the URL-safe base64 alphabet (RFC 4648 §5: `-` for `+`, `_` for `/`), so badge QR artifacts can travel as a URL path segment with no percent-encoding tricks.

## Why

`GET /summits/{id}/badge/{artifact}/validate` takes the artifact as a path segment, and Laravel rawurldecodes the path **before** route matching (`Illuminate/Routing/Matching/UriValidator.php`) with `{badge}` compiled to `[^/]+` — so any artifact whose base64 wrapper contains `/` (near-certain for encrypted badges) can never match the route and 404s, no matter how the client percent-encodes it. Full analysis on ClickUp [86bb7wfgw](https://app.clickup.com/t/9014802374/86bb7wfgw).

Accepting base64url keeps the existing route and payload untouched (one route, one payload — D26): the path segment stays in `[A-Za-z0-9_-]` plus `=` padding, all legal in a path segment.

## How

- `looksLikeBase64`: alphabet regex also accepts `-` and `_`
- `tryBase64Decode`: normalize (`strtr('-_', '+/')`) before the strict `base64_decode`, which rejects the URL-safe alphabet — widening the regex alone is not enough

Additive and backwards-compatible: standard base64 decodes byte-identical; the sole caller is `SummitAttendeeBadge::decodeQRCodeFor`, so badge-scans POST and validateBadge gain it uniformly and existing clients (scanbadgeapp, sponsor lead flows) are unaffected. Checkin is out of scope either way: `AttendeeService::doCheckIn` parses the raw QR directly and never unwrapped base64. Badge QR **emission does not change** — the inner `{hex-IV}{base64(ciphertext)}` layer decoded by `AES::decrypt` stays standard; only the per-request transport wrapper may now be url-safe.

## Tests

`tests/Base64Test.php` (new, pure unit — no app boot): standard alphabet regression, url-safe acceptance, spelling equivalence, padding tolerance, garbage rejection. `OK (5 tests, 11 assertions)`.

## Deployment

Deploy this **before** attendee-networking-api's matching change (its `resolve_badge` re-wraps artifacts url-safe and depends on this decode).

## Summary by CodeRabbit

- **New Features**
  - Added support for decoding URL-safe Base64 values using `-` and `_`.
  - Added automatic padding for valid Base64 values without trailing padding.

- **Bug Fixes**
  - Improved Base64 validation and decoding consistency while preserving standard Base64 support.
  - Invalid, empty, and malformedly padded inputs are now rejected consistently.
  - Equivalent standard and URL-safe encodings decode to the same data.

### DIFF
--- a/app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitApiController.php
+++ b/app/Http/Controllers/Apis/Protected/Summit/OAuth2SummitApiController.php
@@ -2529,7 +2529,7 @@ final class OAuth2SummitApiController extends OAuth2ProtectedController
                 in: 'path',
                 required: true,
                 schema: new OA\Schema(type: 'string'),
-                description: 'RAW Badge QR scan encoded on BASE 64'
+                description: 'RAW Badge QR scan encoded on BASE 64 (standard RFC 4648 §4 or URL-safe §5 alphabet)'
             )
         ],
         responses: [
--- a/app/Utils/Base64.php
+++ b/app/Utils/Base64.php
@@ -19,8 +19,13 @@ final class Base64
         if ($s === '') return false;
         // Standard base64 alphabet (RFC 4648 §4) or URL-safe (§5: '-' and '_'), plus '=' padding
         if (!preg_match('#^[A-Za-z0-9+/_-]*={0,2}$#', $s)) return false;
-        // Length multiple of 4 (padding may be omitted; it gets added below)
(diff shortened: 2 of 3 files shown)


romanetar (approved):
LGTM

PR #577, opened 2026-08-04
fix: gate public presentation serialization on display_on_site for media uploads

JpMaxMan:
ref: https://app.clickup.com/t/9014802374/86bb7zx8t

ref: https://app.clickup.com/t/9014802374/86bb6aem0

## Summary

Public/anonymous callers to the events/published endpoints could pull full
`PresentationMediaUpload` data — including live public S3 URLs — for **draft**
(`display_on_site=false`) uploads via `?expand=media_uploads`, because
`PresentationSerializer` read the unfiltered media uploads collection with no
visibility check.

Reported externally: a third party's calendar/speaker-scraping AI agent recovered
pre-event draft slide decks this way, with no malicious intent — it just added
`expand=media_uploads` to a normal public schedule call.

Closing that hole exposed three further defects in the same serializer, all fixed
here: privileged callers were losing uploads they were entitled to, the response
cache had no audience component, and the cache read had a check-then-act race.

## Root cause

- `GET /summits/{id}/events/published` and `GET /summits/{id}/events/{event_id}/published`
  are fully public/unauthenticated and forward a caller-supplied `expand` with no allowlist.
- `PresentationSerializer::serialize()` handled `expand=media_uploads` (and the
  `relations` id-list path) by calling `Presentation::getMediaUploads()` — unfiltered —
  and serializing every attached upload's `public_url` regardless of draft state.
- `display_on_site` exists on `PresentationMaterial`/`PresentationMediaUpload` for exactly
  this distinction (defaults `false`), and
  `SummitMediaUploadType::getMediaUploadsToDisplayOnSite()` was even built to filter on it —
  but had zero callers in the API/serializer stack.

## What changed

**1. Public callers only see approved uploads** (`90904aa91`)

`getVisibleMediaUploads()` filters to `display_on_site=true` whenever the resolved
serializer type is Public. `AdminPresentationCSVSerializer` is intentionally untouched.

**2. Privileged callers get their uploads back** (`7455f6006`)

`getMediaUploadsSerializerType()` only recognised `isAdmin() || memberCanEdit()`, so two
callers that `OAuth2SummitEventsApiController::getSerializerType()` already treats as
privileged fell through to Public and lost every upload once the filter landed:

- **summit admins** (`summit-front-end-administrators`). The event grid requests
  `media_uploads.display_on_site`, so the operator could no longer see the upload whose
  checkbox is the only thing that would make it visible again — circular.
- **the content-snapshot service account.** pub-api reads with a `client_credentials`
  token, which carries no `user_id`, so `getCurrentUser()` is null by construction.

Service accounts are gated on a **new scope**, `ReadAllPresentationMediaUploads`, rather
than on `ApplicationType_Service` alone, which would have handed drafts to every service
client. The scope is registered with **no endpoint association** on purpose: endpoint
scopes are matched with `array_intersect` (any-of), so associating it would admit a token
holding only this scope to that endpoint.

**3. `media_uploads` is never cached** (`b4079911e`)

Visibility is resolved per user and per scope, but the cache key has no audience
component, so a payload built for one caller could be served to another. Adding the
serializer class to the key does not fix this case — a speaker and a plain attendee both
use `PresentationSerializer`, and a service account with and without the scope both use
`AdminPresentationSerializer`. So the field is stripped before `Cache::put` and recomputed
on every return through a new `withMediaUploads()`, on both the hit and miss paths.

**4. The cache read no longer races** (`e09b148f3`)

`Cache::has()` followed by `Cache::get()` returns null if the entry expires or is evicted
between the two, and `json_decode(null, true)` reaching `withMediaUploads(array $values)`
raised a `TypeError`. One `Cache::get()`; anything that does not decode to an array falls
through and is rebuilt.

**5. The cache key gained a serializer and an unambiguous encoding** (`dea76db68`) —
closes ClickUp 86bb6aem0

`serialize()` is inherited, not overridden, by `AdminPresentationSerializer` and the
track-chair and CSV serializers, and `getAttributeMappings()` merges mappings across the
hierarchy — so the stored payload carried `rank`, `selection_status`, `streaming_url`,
`etherpad_link`, `overflow_stream_key`, chair scores and vote stats. With no class in the
key, a public caller repeating an admin's query params inside the 1200s TTL read that
payload back verbatim.

The key is now `presentation_{id}_{lastEditedTs}_{sha256}`, with the digest over
`static::class`, `expand`, `fields` and `relations`. The parts are `json_encode`d rather
than joined on `_`, which also occurs inside `media_uploads`, `extra_questions` and
`selection_plan` — `expand=media_uploads&fields=x` and `expand=media&fields=uploads_x`
previously rendered the same key. `fields` and `relations` are sorted first; `expand` is
deliberately not, because its relations are dispatched in order and the `speakers` and
`moderator` cases interact through `moderator_speaker_id`.

**6. Serializer-type resolution memoized** (`23a45c253`)

It ran once per media upload plus once per `getVisibleMediaUploads()` — 12 executions for
a ten-upload presentation, measured — and it reaches `memberCanEdit()`, which touches the
member's speaker and this presentation's speaker collection. Memoized **per serializer
instance**, not per request: `memberCanEdit()` is answered against *this* presentation, so
a request-wide memo would hand every presentation in a list the first one's answer.

## Deployment steps

⚠️ **Order matters.** Merging this before steps 1 and 2 empties the content snapshot: the
service account resolves Public, `media_uploads` comes back empty in `events.json` and
`presentations.json`, pub-api does not validate response shape so `SnapshotCompleted` still
fires, and dropbox-materializer stages nothing for every session — silently.

**1. openstackid — register and grant the scope** *(outside this repo)*

Register `{SCOPE_BASE_REALM}/summits/presentations/media-uploads/read/all` on the
summit-api resource server and grant it to the **content-snapshot** client.
`SCOPE_BASE_REALM` is `config('app.scope_base_realm')` per environment, e.g.
`https://api.dev.fnopen.com/summits/presentations/media-uploads/read/all` on dev.

**2. pub-api — request the scope** *(outside this repo)*

Append the same value to `CONTENT_SNAPSHOT_OAUTH2_SCOPES`
(`backend/.env.template:22`, read at `backend/settings.py:321`; space-separated) and
redeploy. Safe to do before summit-api ships — the current code simply ignores the extra
scope on the token.

**3. summit-api — deploy this branch, then run the config migration**

```
php artisan doctrine:migrations:migrate --em=config
```

`Version20260804120000` registers the scope in `api_scopes`. It is idempotent
(`WHERE NOT EXISTS`) and reversible via `down()`. No `endpoint_api_scopes` row is created,
by design — see above. `ApiScopesSeeder` carries the same entry for fresh installs only.

**4. Cache — no action**

The key format change orphans existing entries; they age out on the 1200s TTL and rebuild
on demand. No flush, no coordination.

**5. Verify after deploy**

- Anonymous `?expand=media_uploads` on a published event returns only
  `display_on_site=true` uploads.
- A summit admin loading the summit-admin event grid sees draft uploads again.
- The next content snapshot has non-empty `media_uploads`, and dropbox-materializer stages
  files.

## Test plan

Automated — `tests/PresentationMediaUploadsVisibilityTest.php`,
`tests/PresentationSerializerCacheKeyTest.php`, `tests/PresentationMediaUploadsTest.php`
(13 tests). CI runs them via the new `PresentationMediaUploads` matrix entry in `push.yml`;
the files sit in the `tests/` root, which no existing job covered.

- [x] Public/anonymous, plain attendee, and a service account without the scope see only
      `display_on_site=true` uploads
- [x] Speaker on the presentation, summit admin, and a service account holding the scope see
      drafts too
- [x] An admin-shaped cached payload is not served to a public caller
- [x] Two requests that previously flattened to one key no longer share an entry
- [x] `fields`/`relations` order reuses one entry; `expand` order does not
- [x] An unavailable cached value degrades to a fresh build rather than a 500
- [x] A cache hit is still served, with `media_uploads` resolved fresh
- [x] Manual: summit-admin event grid shows draft uploads after deploy
- [x] Manual: content snapshot non-empty and dropbox-materializer stages files

Each automated test was checked against a deliberately broken implementation, so it is known
to fail when the behaviour regresses rather than assumed to.

`OAuth2PresentationApiTest` is unchanged from `main`: 42 tests, 204 assertions, 6 failures,
2 skipped. Those 6 are pre-existing (`Doctrine\ORM\EntityNotFoundException` on `Member`) and
untouched by this branch.

## Out of scope / follow-up

- This stops the URL from being served going forward — it does **not** revoke the
  public-read ACL already on previously-uploaded draft files. Rotating/regenerating
  already-exposed files (`use_temporary_links_on_public_storage` + the existing
  `presentations-regenerate-media-uploads-temporal-public-urls` command, or a longer-term
  private-storage migration) is a separate ops decision for @smarcet.
- No `display_on_site` backfill. Data audit: 707 rows at `1`, 4514 at `0`; every summit since
  63 is at exactly zero approved. `DocumentsComponent.js:23` filters the same flag client-side,
  so event-site already renders zero media uploads for those shows — the files have always been
  in the payload and never displayed. A blanket backfill would publish material deliberately
  left unapproved on 12/31/63, and on the shows at zero the flag means "never triaged" rather
  than "reviewed and rejected".

## Summary by CodeRabbit

* **New Features**
  * Added controlled access for presentation media uploads, including support for authorized service accounts with the appropriate read-all permission.
  * Public viewers and unauthorized callers see only approved, site-visible uploads.
* **Bug Fixes**
  * Improved presentation response caching so media-upload visibility is evaluated for each request and private data is not inadvertently shared.
  * Cache handling now distinguishes requests with different access levels and expansion options.
* **Tests**
  * Added coverage for media-upload visibility across viewer roles, permissions, and cached responses.

### DIFF
--- a/app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
+++ b/app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
@@ -91,6 +91,26 @@ class PresentationSerializer extends SummitEventSerializer
         return $serializerType;
     }
 
+    /**
+     * Media uploads visible to the resolved serializer type. A Public caller (no admin/editor
+     * privilege on this presentation) only ever sees uploads marked display_on_site=true — an
+     * uploaded-but-not-yet-approved draft (display_on_site=false, the model's own default) must
+     * never reach an unauthenticated/public response, even via ?expand=media_uploads on the
+     * public events/published endpoints.
+     * @return \Doctrine\Common\Collections\Collection|PresentationMediaUpload[]
+     */
+    protected function getVisibleMediaUploads()
+    {
+        $presentation = $this->object;
+        $mediaUploads = $presentation->getMediaUploads();
+        if ($this->getMediaUploadsSerializerType() === SerializerRegistry::SerializerType_Private) {
+            return $mediaUploads;
+        }
+        return $mediaUploads->filter(function ($mediaUpload) {
+            return $mediaUpload->getDisplayOnSite();
+        });
+    }
+
 
     /**
      * @param null $expand
@@ -130,7 +150,7 @@ class PresentationSerializer extends SummitEventSerializer
                         {
                             $media_uploads = [];
 
-                            foreach ($presentation->getMediaUploads() as $mediaUpload) {
+                            foreach ($this->getVisibleMediaUploads() as $mediaUpload) {
                                 $media_uploads[] = SerializerRegistry::getInstance()->getSerializer
                                 (
                                     $mediaUpload, $this->getMediaUploadsSerializerType()
@@ -195,7 +215,7 @@ class PresentationSerializer extends SummitEventSerializer
         if(in_array('media_uploads', $relations))
         {
             $media_uploads = [];
-            foreach ($presentation->getMediaUploads() as $mediaUpload) {
+            foreach ($this->getVisibleMediaUploads() as $mediaUpload) {
                 $media_uploads[] = $mediaUpload->getId();
             }
 
@@ -337,7 +357,7 @@ class PresentationSerializer extends SummitEventSerializer
                     case 'media_uploads':{
                         $media_uploads = [];
 
-                        foreach ($presentation->getMediaUploads() as $mediaUpload) {
+                        foreach ($this->getVisibleMediaUploads() as $mediaUpload) {
                             $media_uploads[] = SerializerRegistry::getInstance()->getSerializer
                             (
                                 $mediaUpload, $this->getMediaUploadsSerializerType()
--- a/app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
+++ b/app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
@@ -224,10 +224,18 @@ class PresentationSerializer extends SummitEventSerializer
 
         $use_cache = $params['use_cache'] ?? false;
 
-        if($use_cache && Cache::has($key)){
-            $values = json_decode(Cache::get($key), true);
-            Log::debug(sprintf("PresentationSerializer::serialize cache hit for presentation %s", $presentation->getId()));
-            return $this->withMediaUploads($values, $expand, $fields, $relations);
+        if($use_cache){
+            // One read, not Cache::has() followed by Cache::get(): the entry can expire on its
+            // own TTL, be evicted, or be flushed in the window between the two, and the second
+            // call would then hand back null for a key the first call vouched for. Anything that
+            // does not decode to an array - a miss, that race, a truncated write - falls through
+            // and is rebuilt.
+            $cached = Cache::get($key);
+            $values = is_string($cached) ? json_decode($cached, true) : null;
+            if(is_array($values)){
+                Log::debug(sprintf("PresentationSerializer::serialize cache hit for presentation %s", $presentation->getId()));
+                return $this->withMediaUploads($values, $expand, $fields, $relations);
+            }
         }
 
         $values = parent::serialize($expand, $fields, $relations, $params);
--- a/app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
+++ b/app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
@@ -209,17 +209,55 @@ class PresentationSerializer extends SummitEventSerializer
         $presentation = $this->object;
         if(!$presentation instanceof Presentation) return [];
 
-        // Include last_edited timestamp so a presentation update naturally busts the cache
-        // without needing an explicit Cache::forget — the old key just ages out via TTL.
+        // fields and relations are read with in_array() throughout, so their order cannot change
+        // the payload and sorting lets two spellings of one request share an entry. $expand is
+        // deliberately NOT sorted: its relations are dispatched in the order given, and the
+        // speakers and moderator cases both write $values['moderator'] while disagreeing about
+        // moderator_speaker_id, so the order is capable of changing the result.
+        $cache_fields = $fields;
+        $cache_relations = $relations;
+        sort($cache_fields);
+        sort($cache_relations);
+
+        // The digest covers everything that shapes the payload and is not already named in the
+        // readable part of the key. static::class is in it because this method is inherited:
+        // AdminPresentationSerializer and the track-chair and CSV serializers all cache through
+        // here, and their merged $array_mappings add fields the public serializer never emits.
+        //
+        // INVARIANT: anything the payload depends on either appears here or stays out of the
+        // cache. static::class covers audience today only because the remaining differences are
+        // class-determined - the mappings are static, getSerializerType() returns a constant per
+        // class, and media_uploads, the one per-user field, is stripped before Cache::put. A
+        // per-user value added to the mappings, or read before parent::serialize() rather than
+        // after it the way AdminPresentationCSVSerializer and TrackChairPresentationSerializer
+        // do, would silently break that.
+        //
+        // json_encode rather than concatenation: the parts contain "_" and "," themselves
+        // (media_uploads, extra_questions, selection_plan), so joining them on those characters
+        // let distinct requests render one key. sha256 rather than md5 because the parts come
+        // from the query string and a collision here means serving one audience's payload to
+        // another - the exact failure the class component is here to prevent.
+        //
+        // last_edited stays readable so a presentation update naturally busts every entry it has
+        // without an explicit Cache::forget, and so an operator can still scan or drop one
+        // presentation's entries by pattern.
         $key =
             sprintf
             (
-                "public_presentation_%s_%s_%s_%s_%s",
+                "presentation_%s_%s_%s",
                 $presentation->getId(),
                 $presentation->getLastEditedUTC()?->getTimestamp() ?? 0,
-                $expand ?? "",
-                implode(",",$fields),
-                implode(",", $relations)
+                hash
+                (
+                    'sha256',
+                    json_encode
+                    ([
+                        'serializer' => static::class,
+                        'expand'     => $expand ?? "",
+                        'fields'     => $cache_fields,
+                        'relations'  => $cache_relations,
+                    ])
+                )
             );
 
         $use_cache = $params['use_cache'] ?? false;
--- a/app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
+++ b/app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
@@ -12,9 +12,11 @@
  * limitations under the License.
  **/
 
+use App\Security\SummitScopes;
 use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Log;
 use Libs\ModelSerializers\AbstractSerializer;
+use models\oauth2\IResourceServerContext;
 use models\summit\Presentation;
 use models\summit\PresentationType;
 
@@ -79,13 +81,40 @@ class PresentationSerializer extends SummitEventSerializer
     ];
 
     /**
+     * Resolves who is allowed to see every media upload attached to this presentation, approved
+     * or not. Kept aligned with OAuth2SummitEventsApiController::getSerializerType(), which is
+     * what decides the serializer type of the presentation itself - when the two disagree the
+     * presentation is served Private while its uploads are served Public, which is the bug this
+     * method used to have for summit admins and for service accounts.
+     *
+     * Two distinct privileged callers:
+     *
+     * - Service accounts (client_credentials, so getCurrentUser() is null by construction) that
+     *   hold the dedicated snapshot scope. The content pipeline stages files pre-event, so it
+     *   needs unapproved uploads. Gated on the scope and not on ApplicationType_Service alone:
+     *   the application type on its own would hand drafts to every service client.
+     * - Members with an admin-level group, or with edit rights over this presentation
+     *   (creator / moderator / speaker).
+     *
      * @return string
      */
     protected function getMediaUploadsSerializerType():string{
+        // && short-circuits, so getCurrentScope() is only reached for service accounts
+        $isSnapshotClient =
+            $this->resource_server_context->getApplicationType() === IResourceServerContext::ApplicationType_Service
+            && in_array
+            (
+                SummitScopes::ReadAllPresentationMediaUploads,
+                $this->resource_server_context->getCurrentScope()
+            );
+
+        if ($isSnapshotClient)
+            return SerializerRegistry::SerializerType_Private;
+
         $serializerType = SerializerRegistry::SerializerType_Public;
         $currentUser = $this->resource_server_context->getCurrentUser();
         $presentation = $this->object;
-        if(!is_null($currentUser) && ( $currentUser->isAdmin() || $presentation->memberCanEdit($currentUser))){
+        if(!is_null($currentUser) && ( $currentUser->isAdmin() || $currentUser->isSummitAdmin() || $presentation->memberCanEdit($currentUser))){
             $serializerType = SerializerRegistry::SerializerType_Private;
         }
         return $serializerType;
--- a/app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
+++ b/app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
@@ -28,6 +28,14 @@ class PresentationSerializer extends SummitEventSerializer
 {
     const CacheTTL = 1200;
 
+    /**
+     * Memo for getMediaUploadsSerializerType(). Instance-scoped on purpose - see that method.
+     * Private rather than protected: the subclasses that override the method answer with a
+     * constant and have nothing to memo.
+     * @var string|null
+     */
+    private ?string $media_uploads_serializer_type = null;
+
     protected static $array_mappings = [
         'CreatorId'               => 'creator_id:json_int',
         'ModeratorId'             => 'moderator_speaker_id:json_int',
@@ -82,10 +90,16 @@ class PresentationSerializer extends SummitEventSerializer
 
     /**
      * Resolves who is allowed to see every media upload attached to this presentation, approved
-     * or not. Kept aligned with OAuth2SummitEventsApiController::getSerializerType(), which is
-     * what decides the serializer type of the presentation itself - when the two disagree the
-     * presentation is served Private while its uploads are served Public, which is the bug this
-     * method used to have for summit admins and for service accounts.
+     * or not. The reference point is OAuth2SummitEventsApiController::getSerializerType(), which
+     * decides the serializer type of the presentation itself - where this method is narrower than
+     * that one the presentation is served Private while its uploads are served Public, which is
+     * the bug it used to have for summit admins and for service accounts.
+     *
+     * It is deliberately still narrower in one place. The controller grants Private to any
+     * ApplicationType_Service caller; here a service account additionally has to hold
+     * ReadAllPresentationMediaUploads, because these are unpublished files and the application
+     * type on its own would hand them to every service client. Members are aligned with the
+     * controller exactly.
      *
      * Two distinct privileged callers:
      *
@@ -99,6 +113,17 @@ class PresentationSerializer extends SummitEventSerializer
      * @return string
      */
     protected function getMediaUploadsSerializerType():string{
+        // Memoized per serializer instance, which is the correct scope and not merely the
+        // convenient one: memberCanEdit() below is answered against THIS presentation, so a
+        // caller can be a speaker on one and a stranger to the next. A request-wide memo would
+        // hand every presentation the first one's answer. SerializerRegistry builds a fresh
+        // serializer per object and none outlive the request, so per-instance already collapses
+        // the repeated work - this method is called once per media upload plus once per
+        // getVisibleMediaUploads(), and it reaches the member's speaker and this presentation's
+        // speaker collection each time.
+        if (!is_null($this->media_uploads_serializer_type))
+            return $this->media_uploads_serializer_type;
+
         // && short-circuits, so getCurrentScope() is only reached for service accounts
         $isSnapshotClient =
             $this->resource_server_context->getApplicationType() === IResourceServerContext::ApplicationType_Service
@@ -109,7 +134,7 @@ class PresentationSerializer extends SummitEventSerializer
             );
 
         if ($isSnapshotClient)
-            return SerializerRegistry::SerializerType_Private;
+            return $this->media_uploads_serializer_type = SerializerRegistry::SerializerType_Private;
 
         $serializerType = SerializerRegistry::SerializerType_Public;
         $currentUser = $this->resource_server_context->getCurrentUser();
@@ -117,7 +142,7 @@ class PresentationSerializer extends SummitEventSerializer
         if(!is_null($currentUser) && ( $currentUser->isAdmin() || $currentUser->isSummitAdmin() || $presentation->memberCanEdit($currentUser))){
             $serializerType = SerializerRegistry::SerializerType_Private;
         }
-        return $serializerType;
+        return $this->media_uploads_serializer_type = $serializerType;
     }
 
     /**
--- a/app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
+++ b/app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
@@ -140,6 +140,62 @@ class PresentationSerializer extends SummitEventSerializer
         });
     }
 
+    /**
+     * Sets media_uploads on an already-built payload for whoever is asking right now, in the
+     * shape the request asked for: an id list for ?relations=media_uploads, serialized objects
+     * for ?expand=media_uploads, expand winning when both are present.
+     *
+     * This is the only place that decides the value, and it runs on every path - including
+     * after a cache read - because getMediaUploadsSerializerType() resolves per user and per
+     * scope, which nothing in the cache key expresses. Two callers can share a key and still
+     * disagree here: a speaker on the presentation and a plain attendee both serialize through
+     * PresentationSerializer, and a service account holding ReadAllPresentationMediaUploads and
+     * one without it both serialize through AdminPresentationSerializer. Adding the serializer
+     * class to the key would not separate either pair.
+     *
+     * @param array $values
+     * @param null $expand
+     * @param array $fields
+     * @param array $relations
+     * @return array
+     */
+    private function withMediaUploads(array $values, $expand, array $fields, array $relations): array
+    {
+        // Nothing asked for it: drop whatever a cached payload may be carrying, so a stale
+        // entry can never contribute this field to a response that did not request it.
+        unset($values['media_uploads']);
+
+        if (in_array('media_uploads', $relations)) {
+            $media_uploads = [];
+            foreach ($this->getVisibleMediaUploads() as $mediaUpload) {
+                $media_uploads[] = $mediaUpload->getId();
+            }
+            $values['media_uploads'] = $media_uploads;
+        }
+
+        if (!empty($expand)) {
+            foreach (explode(',', $expand) as $relation) {
+                if (trim($relation) !== 'media_uploads') continue;
+
+                $media_uploads = [];
+                foreach ($this->getVisibleMediaUploads() as $mediaUpload) {
+                    $media_uploads[] = SerializerRegistry::getInstance()->getSerializer
+                    (
+                        $mediaUpload, $this->getMediaUploadsSerializerType()
+                    )->serialize
+                    (
+                        AbstractSerializer::filterExpandByPrefix($expand, 'media_uploads'),
+                        AbstractSerializer::filterFieldsByPrefix($fields, 'media_uploads'),
+                        AbstractSerializer::filterFieldsByPrefix($relations, 'media_uploads'),
+                    );
+                }
+                $values['media_uploads'] = $media_uploads;
+            }
+        }
+
+        return $values;
+    }
+
 
     /**
      * @param null $expand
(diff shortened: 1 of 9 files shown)


app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
```diff
@@ -91,6 +91,26 @@ protected function getMediaUploadsSerializerType():string{
...
 
+    /**
+     * Media uploads visible to the resolved serializer type. A Public caller (no admin/editor
+     * privilege on this presentation) only ever sees uploads marked display_on_site=true — an
+     * uploaded-but-not-yet-approved draft (display_on_site=false, the model's own default) must
+     * never reach an unauthenticated/public response, even via ?expand=media_uploads on the
+     * public events/published endpoints.
+     * @return \Doctrine\Common\Collections\Collection|PresentationMediaUpload[]
+     */
+    protected function getVisibleMediaUploads()
```
smarcet, line 156: @JpMaxMan This filter is an off switch rather than a draft gate, because `display_on_site` is false on essentially every existing row.

Nothing in the submission flow ever sets it true. `PresentationMaterial::__construct` (PresentationMaterial.php:190) defaults it false, `PresentationMediaUploadFactory::populate` (:39-40) only calls `setDisplayOnSite` when the key is present in the payload, and call-for-presentations - the app speakers actually upload through - has zero occurrences of `display_on_site` in `src/`. The only writer is summit-admin's material form checkbox, which itself starts unchecked (`event-material-reducer.js:30`).

Concrete failure: pub-api's snapshot job requests `expand=media_uploads` with a client_credentials token, which resolves Public here, so every `media_uploads` array in `events.json` and `presentations.json` comes back empty. pub-api does not validate response shape - it only raises on transport errors - so the snapshot publishes normally and `SnapshotCompleted` fires. dropbox-materializer then iterates `sess["media_uploads"]` with no filter of its own and stages nothing, for every session, silently.

Suggested: get a production count first -

```sql
SELECT DisplayOnSite, COUNT(*) FROM PresentationMaterial
WHERE ClassName = 'PresentationMediaUpload' GROUP BY DisplayOnSite;
```

and gate this behind a backfill plus a defined write path for the flag. Also worth a test pinning both branches - the PR's own test plan is still unchecked, and this is exactly the case where behaviour depends entirely on data.

app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
```diff
@@ -91,6 +91,26 @@ protected function getMediaUploadsSerializerType():string{
...
+    {
+        $presentation = $this->object;
+        $mediaUploads = $presentation->getMediaUploads();
+        if ($this->getMediaUploadsSerializerType() === SerializerRegistry::SerializerType_Private) {
```
smarcet, line 160: @JpMaxMan Delegating to `getMediaUploadsSerializerType()` inherits a privilege gap this PR does not close: that method checks only `isAdmin() || memberCanEdit()`, so summit admins and service accounts both land on Public.

Compare `OAuth2SummitEventsApiController::getSerializerType()` (:127), which already resolves Private for `isSummitAdmin()` and for `ApplicationType_Service`. The presentation is served Private to those callers while its media uploads are served Public - the two decisions disagree.

Concrete failure, and it is circular: an operator in `summit-front-end-administrators` loads the summit-admin event grid, which requests `media_uploads.display_on_site` among its fields (`event-actions.js:288`) against `GET /api/v1/summits/{id}/events`. With this filter in place those rows disappear, so the operator cannot see the upload in order to tick the very checkbox that would make it visible again. The same gap is why pub-api's snapshot client gets Public output despite that controller granting service accounts Private.

Suggested: align `getMediaUploadsSerializerType()` with the check the controller already performs - Private for `isSummitAdmin()` as well as `isAdmin()`, and for service accounts. The application-type pattern has serializer-layer precedent at `PresentationSpeakerBaseSerializer.php:104-109`. For the service-account half I would require a dedicated snapshot scope rather than `ApplicationType_Service` alone, so this does not hand drafts to every service client.

app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
```diff
@@ -195,7 +215,7 @@ public function serialize($expand = null, array $fields = [], array $relations =
         if(in_array('media_uploads', $relations))
         {
             $media_uploads = [];
-            foreach ($presentation->getMediaUploads() as $mediaUpload) {
+            foreach ($this->getVisibleMediaUploads() as $mediaUpload) {
```
smarcet: @JpMaxMan This call site never executes when caching is on, so the `relations` id-list path stays unfiltered.

The cached branch above returns at line 170, and on a cache hit it only recomputes `media_uploads` for the `expand` path. A request passing `relations=media_uploads` without `expand=media_uploads` gets the cached id array verbatim - built by whichever caller populated the key, filtered or not.

`use_cache => true` is set on the voteable-presentations endpoints (`OAuth2SummitEventsApiController.php:554`, `:616`, `:685`), which is where pub-api's `presentations.json` feed goes, so this is a live path rather than a theoretical one.

Suggested: handle the `relations` case inside the cache-hit branch the same way `expand` is handled, or drop `media_uploads` from the cached payload entirely and always recompute it after the cache lookup.

app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
```diff
@@ -130,7 +150,7 @@ public function serialize($expand = null, array $fields = [], array $relations =
                         {
                             $media_uploads = [];
 
-                            foreach ($presentation->getMediaUploads() as $mediaUpload) {
+                            foreach ($this->getVisibleMediaUploads() as $mediaUpload) {
```
smarcet: @JpMaxMan Pre-existing rather than introduced here, but it undercuts this PR's premise, so flagging it on the branch you are touching: the cache key is not privilege-scoped.

It is built from presentation id, `last_edited`, `expand`, `fields` and `relations` with no serializer-type or user component, while `getAllVoteablePresentations` passes `$this->getSerializerType()`, which resolves Private or Public per caller. For one combination of params, a payload built by a Private caller can be served to a Public one, and vice versa.

Recomputing `media_uploads` here covers this PR's field, but every other field the Admin serializer contributes still comes straight off the shared key. So "Public callers only ever see approved uploads" is not guaranteed by construction while this stands.

Suggested: add the resolved serializer type to the key. Worth its own ticket rather than expanding this PR's scope.

smarcet (changes requested):
@JpMaxMan please review

app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
smarcet: Update: the counts are in, and they change this recommendation rather than just confirming it.

Global split for `PresentationMediaUpload`: 707 rows at `DisplayOnSite = 1`, 4514 at 0. Per summit, every show since 63 is at exactly zero approved - summit 71, the one in the report and currently live, is 0 visible of 303 uploads across 166 published presentations. The 707 are concentrated in 63 (316), 31 (189) and 12 (169), where the flag was genuinely maintained: summit 63 has 282 hidden uploads but only 18 presentations that lose everything, which is what a working approval flow looks like. So the filter behaves exactly as designed where the practice was kept, and the practice stopped after 63.

I previously suggested gating this behind a backfill. Withdrawing that. There is nothing to preserve: `DocumentsComponent.js:23` filters the same flag client-side, so event-site already renders zero media uploads for summit 71 and every recent show. Those files have always been in the payload and have never been displayed. This is by design, and documented - the internal JSON-contract doc lists `media_uploads` shipping `display_on_site` and `public_url` together, with the filtering deliberately placed in the client.

A blanket backfill is also wrong in both directions: on 12/31/63 it would publish material somebody deliberately left unapproved, and on the shows at zero the flag means "never triaged" rather than "reviewed and rejected", so approval cannot be inferred from it.

Net effect on this PR: no backfill needed, and no user-visible change on event-site. The remaining blockers are the serializer-type alignment and the dropbox-materializer feed, both noted in the other threads.

app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
smarcet: Supporting evidence: track-chairs exercises this path in production. `presentation-actions.js:145-173` sends both `expand=media_uploads,media_uploads.media_upload_type` and `relations=media_uploads` on the same request. It resolves Private via `TrackChairPresentationSerializer.php:59`, so nothing breaks there today, but it confirms the `relations` branch is a live code path rather than a theoretical one.

app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
smarcet: Concrete patch for this, since it is two files and easy to get subtly wrong.

**1) `app/Security/SummitScopes.php`** - new scope, following the existing pattern:

```php
    const ReadSummitData = SCOPE_BASE_REALM.'/summits/read';
    const ReadAllSummitData = SCOPE_BASE_REALM.'/summits/read/all';
    const ReadOverflowEvents = SCOPE_BASE_REALM.'/summits/events/overflow/read';
+   const ReadAllPresentationMediaUploads = SCOPE_BASE_REALM.'/summits/presentations/media-uploads/read/all';
```

**2) `app/ModelSerializers/Summit/Presentation/PresentationSerializer.php`** - imports:

```php
+use App\Security\SummitScopes;
 use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Log;
 use Libs\ModelSerializers\AbstractSerializer;
+use models\oauth2\IResourceServerContext;
 use models\summit\Presentation;
 use models\summit\PresentationType;
```

**3) Same file** - replace `getMediaUploadsSerializerType()`:

```php
    protected function getMediaUploadsSerializerType(): string
    {
        $currentUser  = $this->resource_server_context->getCurrentUser();
        $presentation = $this->object;

        // Service account (client_credentials => no current user) holding the dedicated
        // snapshot scope. The content pipeline needs unapproved uploads in order to stage
        // files pre-event. Narrowed by scope on purpose: ApplicationType_Service on its own
        // would hand drafts to every service client.
        $isSnapshotClient =
            $this->resource_server_context->getApplicationType() === IResourceServerContext::ApplicationType_Service
            && in_array(
                SummitScopes::ReadAllPresentationMediaUploads,
                $this->resource_server_context->getCurrentScope()
            );

        if ($isSnapshotClient)
            return SerializerRegistry::SerializerType_Private;

        if (!is_null($currentUser)
            && ($currentUser->isAdmin()
                || $currentUser->isSummitAdmin()      // <- show admin operators, currently missing
                || $presentation->memberCanEdit($currentUser))) {
            return SerializerRegistry::SerializerType_Private;
        }

        return SerializerRegistry::SerializerType_Public;
    }
```

Notes on the shape:

- `getVisibleMediaUploads()` needs no change. It already keys off this method, which is why that hook was the right choice.
- `TrackChairPresentationSerializer.php:59` overrides this and returns Private unconditionally, so track chairs are unaffected either way.
- The `in_array` scope check follows the precedent already in the serializer layer at `SummitSerializer.php:372-377`, and `getCurrentScope()` returns an array (see `OAuth2SummitEventsApiController.php:125`). The `&&` short-circuits, so `getCurrentScope()` is only reached for service accounts.
- `isSummitAdmin()` resolves `summit-front-end-administrators` (`Member::isSummitAdmin`), which is the group the show-admin operators are in and the reason the grid goes blank today.

Out-of-code step, easy to forget: the new scope has to be provisioned on the content-snapshot client in the IDP, and added to pub-api's `CONTENT_SNAPSHOT_OAUTH2_SCOPES`. Until both are done pub-api keeps resolving Public and the snapshot still empties.

Worth a test per branch on this method: anonymous, attendee, speaker on the presentation, summit admin, service account with the scope, service account without it. Six cases, all cheap, and they pin the exact behaviour this PR depends on.

smarcet: Follow-up on the scope: it needs a seeder entry and a migration, and the answer to "which endpoint does it attach to" is **none**. Worth spelling out because the obvious precedent points the wrong way.

**Why no endpoint association.** `OAuth2BearerAccessTokenRequestValidator.php:193` rejects with `count(array_intersect($endpoint_scopes, $token_scopes)) == 0`, so endpoint scopes are any-of, not all-of. Associating this scope with `/summits/{id}/events/published` would mean a token holding *only* this scope is admitted to that endpoint - widening access rather than narrowing it. And it is unnecessary: `getMediaUploadsSerializerType()` reads `getCurrentScope()` off the token directly and never consults `endpoint_api_scopes`.

That is the difference from `ReadOverflowEvents` (`Version20260715120000.php`), which was minted alongside a brand-new endpoint it gates. This one is a privilege modifier over endpoints that already exist.

**Two registries, only one of which is load-bearing.**

- summit-api `config` DB (`apis` / `api_scopes` / `endpoint_api_scopes`): validates that an incoming token intersects the endpoint's scopes. Registering the scope here is convention and discoverability (Swagger `security` annotations), not a functional requirement for this check.
- openstackid (`oauth2_api_scope`): issues the token. **This is the one that matters.** If the scope does not exist there and is not granted to the content-snapshot client, the token never carries it and the serializer never sees it. `openstackid/database/seeds/ApiScopeSeeder.php` only seeds the IDP's own APIs - no summit scopes - so this is administered through the IDP admin UI, an ops step rather than a code change.

**1) `database/seeders/ApiScopesSeeder.php`**, inside `seedSummitScopes()`, next to `ReadOverflowEvents`:

```php
            [
                'name' => SummitScopes::ReadAllPresentationMediaUploads,
                'short_description' => 'Read All Presentation Media Uploads',
                'description' => 'Grants read access to presentation media uploads regardless of display_on_site, for trusted service accounts feeding the content pipeline',
            ],
```

**2) `database/migrations/config/Version20260804120000.php`** (latest currently is `Version20260715120000`):

```php
<?php namespace Database\Migrations\Config;

use App\Security\SummitScopes;
use Doctrine\DBAL\Schema\Schema;
use Doctrine\Migrations\AbstractMigration;

/**
 * Register the ReadAllPresentationMediaUploads scope.
 *
 * Adds 1 api_scopes row and NO endpoint association, on purpose: endpoint scopes are
 * validated with array_intersect (any-of, see OAuth2BearerAccessTokenRequestValidator:193),
 * so associating this scope with an existing endpoint would admit a token holding only
 * this scope. The scope is read directly off the token by
 * PresentationSerializer::getMediaUploadsSerializerType(), which never consults
 * endpoint_api_scopes.
 *
 * Idempotent via WHERE NOT EXISTS.
 */
final class Version20260804120000 extends AbstractMigration
{
    use APIEndpointsMigrationHelper;

    private const API_NAME = 'summits';

    public function getDescription(): string
    {
        return 'Register ReadAllPresentationMediaUploads scope (no endpoint association).';
    }

    public function up(Schema $schema): void
    {
        $this->addSql($this->insertApiScope(
            self::API_NAME,
            SummitScopes::ReadAllPresentationMediaUploads,
            'Read All Presentation Media Uploads',
            'Grants read access to presentation media uploads regardless of display_on_site, for trusted service accounts feeding the content pipeline'
        ));
    }

    public function down(Schema $schema): void
    {
        $this->addSql($this->deleteApiScopes(self::API_NAME, [SummitScopes::ReadAllPresentationMediaUploads]));
    }
}
```

`insertApiScope()` is independent of `registerEndpoint()` in `APIEndpointsMigrationHelper`, so a scope with no endpoint association is a supported shape rather than a workaround.

Rollout order matters: IDP scope + client grant first, then pub-api's `CONTENT_SNAPSHOT_OAUTH2_SCOPES`, then this PR. Land the serializer filter before the client actually holds the scope and the snapshot empties in between.

smarcet: @JpMaxMan Resolved in `7455f6006`, with the six-branch test coverage this thread asked for landing in `9af1aed07`.

`getMediaUploadsSerializerType()` (`PresentationSerializer.php:101`) now recognises two privileged callers rather than one-and-a-half: the snapshot service account, gated on `ApplicationType_Service` **and** `ReadAllPresentationMediaUploads` being on the token so the application type alone does not admit every service client; and members who are global admins, summit admins via `isSummitAdmin()`, or presentation editors via `memberCanEdit()`. That closes the circular failure — the `summit-front-end-administrators` operator can see the upload again in the event grid, which is the only place the `display_on_site` checkbox lives.

The scope is wired in three places: `SummitScopes.php:25`, `ApiScopesSeeder.php:72`, and migration `Version20260804120000.php`, which registers it with **no** `endpoint_api_scopes` association, for the `array_intersect` reason spelled out in the follow-up above.

Tests are `tests/PresentationMediaUploadsVisibilityTest.php` — exactly the six cases named here: anonymous, plain attendee, speaker on the presentation, summit admin, service account with the scope, service account without it. They assert on the media upload ids `serialize()` returns rather than on the serializer-type string, so a decision reached another way would not break them. Each was checked against a deliberately broken implementation rather than assumed to bite: dropping `isSummitAdmin()` fails the summit-admin case alone, and gating on `ApplicationType_Service` without the scope check fails the without-scope case alone.

Still open, outside this repo, and it should gate the rollout rather than the merge: the scope has to be provisioned on the content-snapshot client in the IDP and added to pub-api's `CONTENT_SNAPSHOT_OAUTH2_SCOPES`. Until both are done that client still resolves Public and the snapshot still empties.

app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
smarcet: @JpMaxMan Resolved in `b4079911e`, taking the second of the two options this thread offered: `media_uploads` is dropped from the cached payload entirely and recomputed after the lookup, rather than the `relations` case being duplicated inside the cache-hit branch.

The reason for preferring it is the one from the sibling thread on the cache key — adding the serializer class to the key would not separate the pairs that actually disagree. A speaker on the presentation and a plain attendee both serialize through `PresentationSerializer`; a service account holding `ReadAllPresentationMediaUploads` and one without it both serialize through `AdminPresentationSerializer`. As long as the field is stored at all, correctness depends on every future reader of this method remembering to recompute it.

Shape of the fix: `withMediaUploads()` (`PresentationSerializer.php:162`) is now the only place that sets the field, in the shape the request asked for — id list for `relations`, serialized objects for `expand`, `expand` winning when both are present, which is what track-chairs sends. It runs on the cache-hit path (`:237`) and on the return of the non-cached path (`:474`), and `Cache::put` receives a copy with the field removed (`:469-471`). It opens by `unset()`ing the field, so an entry written before this change cannot leak one either as it ages out.

`e09b148f3` adds the regression test for the guarantee that matters here: returning the cached `$values` without recomputing lets the stale draft upload in the stored payload reach a public caller, and the test fails on exactly that.

Worth noting this narrows the field without privilege-scoping the key itself — every other Admin-serializer field still comes off the shared key. That is the sibling thread's subject and still wants its own ticket.

app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
```diff
@@ -122,32 +227,7 @@ public function serialize($expand = null, array $fields = [], array $relations =
...
-                                );
-                            }
-
-                            $values['media_uploads'] = $media_uploads;
-                        }
-                    }
-                }
-            }
-            return $values;
+            return $this->withMediaUploads($values, $expand, $fields, $relations);
```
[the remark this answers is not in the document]
smarcet: Correct, and it was made worse by the commit under review rather than merely surviving it. `json_decode(null, true)` is `null`, and once the cache-hit branch started calling `withMediaUploads(array $values, ...)` that `null` became a `TypeError` — a 500 where the previous code returned a quietly wrong payload.

Fixed in `e09b148f3`, with the check-then-act pattern removed rather than the null handled: a single `Cache::get()`, and anything that does not decode to an array falls through to the normal build path. That covers the eviction race, an ordinary miss, and a truncated write on one route instead of three.

The regression test you asked for is `testUnavailableCachedValueIsTreatedAsAMiss` in `tests/PresentationMediaUploadsVisibilityTest.php`. It asserts on the resulting payload rather than on how the cache was consulted, so it does not need rewriting if that read changes shape again, and it fails with exactly the `TypeError` above against the previous code.

app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
smarcet: @JpMaxMan Closed in `dea76db68`, so this no longer outlives the branch.

The key is now `presentation_{id}_{lastEditedTs}_{sha256}`, with the digest over `static::class`, `expand`, `fields` and `relations` (`PresentationSerializer.php:244-261`). That is the discriminator this thread asked for: `serialize()` is inherited rather than overridden by `AdminPresentationSerializer`, `TrackChairPresentationSerializer` and `AdminPresentationCSVSerializer`, so the class is what separates their merged mappings.

Two things beyond what this thread scoped, both found while implementing it:

The parts were being joined with `_`, which also occurs inside `media_uploads`, `extra_questions` and `selection_plan` — so `expand=media_uploads&fields=x` and `expand=media&fields=uploads_x` rendered the same key regardless of the class component. `json_encode` is what closes that; hashing the old concatenation preserves it exactly, and there is a test asserting that. `sha256` over `md5` since these parts are caller-controlled and a collision means serving one audience's payload to another.

`fields` and `relations` are now sorted before hashing — both are read with `in_array()`, so order cannot change the payload. `expand` is left alone on purpose, because the `speakers` and `moderator` cases both write `$values['moderator']` while disagreeing about `moderator_speaker_id`; a test pins that so it does not get normalised later.

Tests are `tests/PresentationSerializerCacheKeyTest.php`, four cases, the first three written failing first. Each was also checked against a broken implementation rather than assumed to bite.

The invariant this rests on is now a comment on the key rather than tribal knowledge: audience-dependent data either appears in the key or stays out of the cache. `static::class` is enough only while the remaining differences are class-determined — mappings are static, `getSerializerType()` is constant per class, and `media_uploads` is stripped before `Cache::put` (`:503-509`). Notably `AdminPresentationCSVSerializer` and `TrackChairPresentationSerializer` do read the current user, but both call `parent::serialize()` first and shape afterwards, so their per-user work never reaches the cache. That ordering was undocumented and is exactly what a future change could break without noticing.

ClickUp `86bb6aem0` has the full write-up and can be closed.

app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
```diff
@@ -79,18 +81,121 @@ class PresentationSerializer extends SummitEventSerializer
     ];
 
     /**
+     * Resolves who is allowed to see every media upload attached to this presentation, approved
+     * or not. Kept aligned with OAuth2SummitEventsApiController::getSerializerType(), which is
+     * what decides the serializer type of the presentation itself - when the two disagree the
+     * presentation is served Private while its uploads are served Public, which is the bug this
+     * method used to have for summit admins and for service accounts.
+     *
```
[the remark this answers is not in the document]
smarcet: Fixed in `23a45c253`. The docblock now states the divergence in its opening rather than three paragraphs down: the controller grants Private to any `ApplicationType_Service` caller, while this method additionally requires `ReadAllPresentationMediaUploads`, because these are unpublished files and the application type alone would hand them to every service client. Members remain aligned with the controller exactly.

Worth noting the qualification was already present further down — the problem was that the first sentence claimed full alignment, so a reader who stopped there would conclude the opposite of what the code does. That is the part that was worth fixing.

app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
```diff
@@ -79,18 +81,121 @@ class PresentationSerializer extends SummitEventSerializer
...
+                    $media_uploads[] = SerializerRegistry::getInstance()->getSerializer
+                    (
+                        $mediaUpload, $this->getMediaUploadsSerializerType()
+                    )->serialize
+                    (
+                        AbstractSerializer::filterExpandByPrefix($expand, 'media_uploads'),
+                        AbstractSerializer::filterFieldsByPrefix($fields, 'media_uploads'),
+                        AbstractSerializer::filterFieldsByPrefix($relations, 'media_uploads'),
+                    );
+                }
```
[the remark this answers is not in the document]
smarcet: Fixed in `23a45c253`, though the memo went on the method rather than on the call site, which turns out to matter: hoisting the result out of the loop in `withMediaUploads()` would have left the call `getVisibleMediaUploads()` makes on its own. Measured on a ten-upload presentation serialized with both `expand=media_uploads` and `relations=media_uploads`: 12 executions of the method body before, 1 after.

One correction on the reasoning. Current-user resolution is not part of the cost — `ResourceServerContext::getCurrentUser()` is already memoized per request behind an explicit fast path, and the voteable endpoints resolve it in the controller before serialization begins, so inside the loop it is a property read. The real repeated work is `Presentation::memberCanEdit()`, which reaches the member's speaker and this presentation's speaker collection.

The memo is scoped to the serializer instance rather than the request, and that distinction is load-bearing: `memberCanEdit()` is answered against *this* presentation, so a caller can be a speaker on one and a stranger to the next. A request-wide memo would hand every presentation in a list the first one's answer — the same audience-confusion failure this PR exists to close. `SerializerRegistry` builds a fresh serializer per object and none outlive the request, so per-instance is both correct and sufficient.

app/ModelSerializers/Summit/Presentation/PresentationSerializer.php
```diff
@@ -104,50 +234,71 @@ public function serialize($expand = null, array $fields = [], array $relations =
...
+                (
+                    'sha256',
+                    json_encode
+                    ([
+                        'serializer' => static::class,
+                        'expand'     => $expand ?? "",
+                        'fields'     => $cache_fields,
+                        'relations'  => $cache_relations,
+                    ])
+                )
```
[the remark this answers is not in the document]
smarcet: @JpMaxMan Fixed in `7106979ee`, taking the third of the three options this comment offered: the `json_encode()` result is captured, and a request whose parts cannot be encoded bypasses the cache entirely — no read, no write, served fresh. `JSON_INVALID_UTF8_SUBSTITUTE` was considered and rejected because substitution maps every invalid sequence to the same U+FFFD, so two requests differing only in *which* malformed byte they carried would still share an entry; rejecting the request outright would put an HTTP-level decision inside a serializer that has no business making one.

The finding reproduced end to end, not just in theory: `Illuminate\Http\Request::create("...?expand=media%FFuploads")` hands `input('expand')` the raw `0xFF` byte, `json_encode` of the exact digest structure returns `false`, and `hash('sha256', false)` yields `sha256("")` with no warning even at `E_ALL` — the file does not declare `strict_types`. Two requests with different serializer class, expand, fields and relations produced the identical digest, which voids the class discriminator `dea76db68` added.

One caveat on severity, for the record: the cross-audience direction needs a *privileged* caller to emit malformed UTF-8 inside the TTL, which nothing in the platform does organically and an attacker cannot force — so this is hardening rather than a live hole. It is closed anyway because the digest's whole contract is that a collision never crosses audiences.

Regression pinned by `testMalformedUtf8RequestsDoNotCollideOnOneEntry` (`PresentationSerializerCacheKeyTest.php`): before the guard, an admin-shaped payload cached under the collapsed digest came back verbatim to a public caller whose request also failed to encode; the test asserts it no longer does. All 14 tests in the three suite files pass.

smarcet (approved):
LGTM

romanetar (approved):
LGTM

PR #581, opened 2026-08-08
feat: add per-activity CFP reopen, an admin-set time-boxed submission override

caseylocker:
ref: https://app.clickup.com/t/86bba82ch

Spec: `sds/per-activity-cfp-reopen.md` in the ftn-docsnsklz vault, on `main` (merged `bd2c2c3`), sections 3, 4 and 7.

## What this does

A summit admin can reopen CFP submission for a single presentation for a chosen window (default 24h, ceiling 168h), so the speaker can edit that one talk through the existing submission flow after the selection plan's window has closed. Expiry is passive: no cron, no cleanup job.

This is step 1 of 3. The call-for-presentations and summit-admin work are ClickUp `86bba82ph` and `86bba82y3`. Until step 2 ships, nothing a speaker can see changes.

## How it works

Three nullable columns on `Presentation` store the grant: raw hours, stamp date, granting member. **The deadline is derived on read and never stored**, so the granted duration and the window end cannot drift out of lockstep.

A new `isSubmissionReopened()` predicate gates on that derived deadline plus plan invariants (assigned plan, enabled, submission window already ended). It is folded into `isSubmissionClosed()` as an early return, which covers both delete guards with no change at either call site, and it is OR'd into the two `isSubmissionOpen()` checks on update and complete. Two admin-only endpoints stamp and clear the grant.

The plan invariants are not optional. A deadline-only check would grant edits before the CFP ever opened, and via the `isSubmissionClosed()` fold it would open speaker deletes on a plan disabled after the grant, because the delete path never re-checks `IsEnabled()` itself.

## Notes for reviewers

Seven things in this diff look like defects but are deliberate. Each cost a review round to establish, so they are recorded here rather than rediscovered.

1. **Both routes carry `auth.user`, and both endpoint registrations carry authz groups**: `SuperAdmins`, `Administrators`, `SummitAdministrators`, mirroring `update-event` minus the track chair roles. The middleware is a coarse global group check and the summit scoped admin check still runs in the controller. It narrows nothing, because `Member::hasPermissionForOnGroup()` already ends in `&& isOnGroup($slug)`, so every caller that passes the controller check passes the middleware; a client credentials token with no member passes the middleware and is still refused by the controller. The groups are what make the middleware usable at all: `UserAuthEndpoint::handle()` returns 403 when an endpoint's required group list is empty.

2. **Serialization uses `SerializerType_Private`, not `_Admin`.** Presentation has no `Admin` key, and an unknown type silently falls back to Public, which would strip every field these endpoints exist to set.

3. **Scopes are the OR trio `WriteSummitData`, `WriteEventData`, `WritePresentationData`.** Validation is any of, so the trio admits existing Show Admin tokens. Narrowing to `WritePresentationData` alone would 403 every current caller.

4. **`submission_reopened_by` is an expandable relation, declared on `AdminPresentationSerializer` rather than on the base.** The payload carries `submission_reopened_by_id`, and `?expand=submission_reopened_by` replaces it with the serialized member. The mapping is deliberately subclass local: `AbstractSerializer::getExpandsMappings()` merges lineage parent to child only, and `SubmissionPresentationSerializer` is a sibling of `AdminPresentationSerializer`, so the relation is unreachable from the Submission and Public variants. A case in the base expand switch would not be, which is the leak the Admin-only design prevents and the reason the SDS rejected id plus expand when a base class switch was the only mechanism considered. The expand serializes the member as Private, matching how `created_by` is already serialized on this class. Show Admin therefore asks for the expand; that ticket is not yet started, so nothing downstream changes.

5. **The two speaker presentation-list endpoints previously serialized Public and now serialize Submission**, which also selects Admin serializers for the nested `created_by`, `updated_by`, `moderator` and `speakers` relations. This grants no new access: `getPresentationSubmission` already serializes the same fields for the same audience, gated by the same `memberCanEdit` check (creator, moderator, or assigned speaker). The switch itself is what the SDS mandates, since `PagingResponse::toArray()` defaults to Public and the CFP portal table would otherwise never receive the new field.

6. **The service's `$hours < 1` check is unreachable over HTTP**, because the endpoint validates `sometimes|integer|min:1` and refuses first. It is kept deliberately: it is the only guard for a non HTTP caller, and a persisted non positive value would make `getSubmissionReopenedUntil()` throw on every read, since `new \DateInterval('PT-1H')` is invalid.

7. **A misconfigured default is clamped, while an explicit out of range value is refused.** If `default_reopen_hours` is configured above `max_reopen_hours`, the resolved default is clamped down to the ceiling rather than rejected. An explicitly supplied `hours` above the ceiling still gets a 412. The asymmetry is deliberate: a configuration error the caller can neither see nor fix should not surface to an admin as an unexplainable refusal of a request that supplied nothing, whereas a caller who names an out of range value should be told. Covered by `testMisconfiguredDefaultAboveMaxIsClampedRatherThanRefused`.

Two further points on the design:

- **The model schema change is split across two migrations, mirroring `Version2026061500000{0,1}`.** Within one migration every `addSql()` statement runs before the `Builder` schema diff, so a single migration doing both would execute the foreign key against a column that did not exist yet. `Version20260807120000` carries the three nullable columns and the index, entirely through `Builder`. `Version20260807120001` carries the foreign key as its single `addSql()` statement and only reads the schema, so it emits no diff and the hazard cannot reappear. Both guard per component, so an interrupted run repairs on retry. Confirmed against a scratch database: the resulting schema has one index named `SubmissionReopenedByID` and unchanged column types.
- **Both the seeder entry and the config migration are needed and are not redundant.** The deploy flow does not re-run seeders, and an unregistered route returns 400 before the controller runs.

## Deploy notes

1. **The foreign key addition copies the table.** On MySQL 8.0 an in place foreign key addition requires `foreign_key_checks` to be disabled; otherwise it rebuilds. `Presentation` is a large production table, so run this migration in a low traffic window and monitor lock time. The three nullable columns are combined into a single `ALTER`, and the explicit index on the FK column is the index MySQL would create anyway.

2. **A model migration that fails partway is safe to re-run.** The DDL statements commit separately, so a failure at the index or foreign key step leaves the added columns in place with the migration unrecorded. Both model migrations guard each component on itself rather than on the column, so a retry adds only what is missing instead of failing on duplicate schema objects.

3. **Migrations use `--em=config` and `--em=model_write`.** There is no `model` entity manager; `--em=model` silently finds zero migrations and reports "already at latest".

4. **Safe rollback is application only.** The columns are additive and inert when null. Rolling the schema back destroys live grants and the actor and date audit trail, so prefer rolling back the application. If a schema rollback is unavoidable, confirm all serving instances run the old application first.

5. **Two new optional config values, and no `.env` change is required to deploy.** `CFP_MAX_REOPEN_HOURS` (ceiling on an admin granted window, default 168, seven days) and `CFP_DEFAULT_REOPEN_HOURS` (window used when the caller sends no `hours`, default 24). Both are documented in `.env.example`, and both defaults live in `config/cfp.php`, so an environment that sets neither behaves as specified. Set them per environment only if a show needs a different ceiling. A `CFP_DEFAULT_REOPEN_HOURS` configured above the ceiling is clamped down rather than rejected, per reviewer note 7 above, so a typo there raises no error anywhere.

    On k8s specifically: neither key has to be added to the configmap for this to deploy, because the fallbacks in `config/cfp.php` are the intended values. They only need adding where an environment wants a non-default ceiling or default. If they are added, they have to be present in the pod environment **before the container's entrypoint runs `php artisan config:cache`**, which `fn-docker/summit-api/php-entry-point.sh` does on every start: `config/cfp.php` resolves `env()` at cache build time, so a value injected into the pod after that point is baked out and reads back as the default.

6. **Run both migrations with or before the application deploy.** The entity maps the three new columns, which only `Version20260807120000` and `Version20260807120001` (`--em=model_write`) create, and the routes read their authz groups from the config database, which only `Version20260807130000` (`--em=config`) populates. Application code live ahead of either one fails: presentation reads hit missing columns, and the reopen routes answer 400 for an unregistered endpoint.

## Endpoint registration evidence

Against the config database, before the migration:

```sql
SELECT name FROM api_endpoints WHERE name LIKE '%presentation-submission-period%';
-- 0 rows
```

After `php artisan doctrine:migrations:migrate --em=config`:

```
reopen-presentation-submission-period   PUT     /api/v1/summits/{id}/presentations/{presentation_id}/submission-period/reopen
close-presentation-submission-period    DELETE  /api/v1/summits/{id}/presentations/{presentation_id}/submission-period/reopen
-- 2 rows, each carrying summits/write, summits/write-event, summits/write-presentation
-- each also carrying authz_groups super-admins, administrators, summit-front-end-administrators
```

Both route strings byte match the Laravel routes as printed by `route:list`.

## Testing

97 new tests across five files, 45 integration and 52 unit:

| File | Tests | Kind | Covers |
|---|---|---|---|
| `tests/PresentationReopenModelTest.php` | 8 | integration | predicates, derived deadline, the `isSubmissionClosed()` fold |
| `tests/PresentationReopenApiTest.php` | 25 | integration | endpoints, serialization, CFP table feeds, acceptance under an active grant, open-window parity |
| `tests/PresentationReopenAuthzTest.php` | 12 | integration | non-admin 403s, both delete paths, authorship, the create gate |
| `tests/Unit/Models/PresentationSubmissionReopenTest.php` | 29 | unit | the full deadline and predicate matrix, including half-states and boundaries |
| `tests/Unit/Services/PresentationSubmissionReopenServiceTest.php` | 23 | unit | every validation branch of the reopen service |

Integration shard: `OK (59 tests, 385 assertions)`. Unit files: `OK (29 tests, 59 assertions)` and `OK (23 tests, 77 assertions)`.

**All five files are registered in the `push.yml` matrix.** No CI job covers the `tests/` root, and `tests/Unit/Models/` was not previously sharded either, so a test in either location runs nowhere unless it is named.

The two unit files are genuinely database-free and framework-free: they extend plain `PHPUnit\Framework\TestCase`, bind a real config repository into a bare container for the facade, and mock `ITransactionService` so the transaction callback still executes. **52 tests run in about 120ms**, against roughly 50 seconds for the integration shard. They exist because several cases are impractical or impossible to reach otherwise: the expiry boundary, the half-states where hours is set with a null date or the reverse, and the service's `$hours < 1` guard, which the endpoint's `min:1` rule makes unreachable over HTTP so the unit test is its only coverage.

Two notes on test rigour, both of which changed how these were written:

- **The authz tests run against a genuinely non-admin identity.** `ProtectedApiTestCase` produces a global admin through two independent levers, the persisted group and the access token stub's default IdP groups, and both delete guards short circuit for a global admin. Both levers are defeated and a canary asserts it, because an admin run 403 suite proves nothing.
- **Several tests arrange their precondition directly on the model rather than through the reopen endpoint.** A BrowserKit test cannot perform two sequential HTTP writes against the same entity: `DoctrineMiddleware` closes the model entity manager after every request and singleton repositories pin the manager they were first resolved with, so the second write is silently dropped while still returning success. Each affected test asserts its precondition landed in the database before issuing the request under test.

The full suite is green in CI, including the `tests/oauth2/` shard at 1069 tests. Local full-suite runs need two setup steps that CI does for itself: freshly created test databases (`php artisan db:create_initial_db` for both schemas, as `push.yml` does), and `max_connections` above MySQL's default of 151, which the suite exceeds. Without those, a local run reports failures that are entirely environmental.

## Not in scope

- Cache key work, already shipped in #577.
- The delete bypass role narrowness, ClickUp `86bba832v`.
- A window guard on media and speaker subresource mutations, ClickUp `86bba8388`.
- Presentation creation under a closed plan. A test asserts creation stays blocked.

Three pre-existing defects were found while implementing this and deliberately left untouched, so they are not mistaken for regressions or for oversights:

- `SelectionPlan::areFieldsEqual` compares its first argument to itself, so the scalar branch always reports equal and the allowed editable question guard never fires for scalar fields.
- `PresentationType::isAreSpeakersMandatory()` ignores the `are_speakers_mandatory` column and returns `min_speakers > 0`, so calling the setter has no effect.
- `getPresentationMediaUploads` is registered twice in `routes/api_v1.php`.

## Scope compatibility with Show Admin, checked

Scope validation is any of, and the endpoints accept `WriteSummitData` (`summits/write`), `WriteEventData` (`summits/write-event`) or `WritePresentationData` (`summits/write-presentation`). summit-admin's requested scopes, per its `.env.example`, are:

```
summits/read
summits/read/all
summits/write                          <- WriteSummitData
summits/write-event                    <- WriteEventData
summits/write-presentation-materials   <- note: NOT write-presentation
```

Two of the three, so Show Admin is admitted without any client change.

This is also why the trio matters rather than being belt and braces. summit-admin holds `write-presentation-materials`, not `write-presentation`, so registering `WritePresentationData` alone would have returned 403 to every Show Admin caller in production while passing every test here, because the test token does carry it.

One residual check for whoever deploys: `.env.example` is the template, and a deployed `.env` can drift from it. Worth confirming the deployed summit-admin `.env` still lists `summits/write` before the control in `86bba82y3` goes live.

## Summary by CodeRabbit

* **New Features**
  * Administrators can reopen closed presentation submission periods for a configurable duration.
  * Administrators can immediately close reopened submission periods.
  * Reopening details, deadlines, and administrator information are visible where applicable.
  * Configurable default and maximum reopening windows are now supported.

* **Bug Fixes**
  * Validly reopened presentations can be updated and completed during their reopening window, while existing permissions remain enforced.

* **Tests**
  * Added comprehensive coverage for reopening, closing, authorization, validation, serialization, and configuration behavior.

### DIFF
--- a/app/Http/Controllers/Apis/Protected/Summit/OAuth2PresentationApiController.php
+++ b/app/Http/Controllers/Apis/Protected/Summit/OAuth2PresentationApiController.php
@@ -593,7 +593,11 @@ final class OAuth2PresentationApiController extends OAuth2ProtectedController
             // unknown type silently falls back to Public, stripping the reopen fields.
             return $this->updated(SerializerRegistry::getInstance()->getSerializer(
                 $presentation, SerializerRegistry::SerializerType_Private
-            )->serialize());
+            )->serialize(
+                SerializerUtils::getExpand(),
+                SerializerUtils::getFields(),
+                SerializerUtils::getRelations()
+            ));
         });
     }
 
--- a/app/Http/Controllers/Apis/Protected/Summit/OAuth2PresentationApiController.php
+++ b/app/Http/Controllers/Apis/Protected/Summit/OAuth2PresentationApiController.php
@@ -43,6 +43,7 @@ use ModelSerializers\IPresentationSerializerTypes;
 use ModelSerializers\SerializerRegistry;
 use OpenApi\Attributes as OA;
 use services\model\IPresentationService;
+use services\model\IPresentationSubmissionReopenService;
 use utils\Filter;
 use utils\FilterElement;
 use utils\FilterParser;
@@ -85,6 +86,11 @@ final class OAuth2PresentationApiController extends OAuth2ProtectedController
     private $presentation_comments_repository;
 
     /**
+     * @var IPresentationSubmissionReopenService
+     */
+    private $presentation_submission_reopen_service;
+
+    /**
      * OAuth2PresentationApiController constructor.
      * @param IPresentationService $presentation_service
      * @param ISummitRepository $summit_repository
@@ -92,6 +98,7 @@ final class OAuth2PresentationApiController extends OAuth2ProtectedController
      * @param IMemberRepository $member_repository
      * @param ISummitPresentationCommentRepository $presentation_comments_repository
      * @param IResourceServerContext $resource_server_context
+     * @param IPresentationSubmissionReopenService $presentation_submission_reopen_service
      */
     public function __construct
     (
@@ -100,7 +107,8 @@ final class OAuth2PresentationApiController extends OAuth2ProtectedController
         ISummitEventRepository               $presentation_repository,
         IMemberRepository                    $member_repository,
         ISummitPresentationCommentRepository $presentation_comments_repository,
-        IResourceServerContext               $resource_server_context
+        IResourceServerContext               $resource_server_context,
+        IPresentationSubmissionReopenService  $presentation_submission_reopen_service
     )
     {
         parent::__construct($resource_server_context);
@@ -109,6 +117,7 @@ final class OAuth2PresentationApiController extends OAuth2ProtectedController
         $this->member_repository = $member_repository;
         $this->summit_repository = $summit_repository;
         $this->presentation_comments_repository = $presentation_comments_repository;
+        $this->presentation_submission_reopen_service = $presentation_submission_reopen_service;
     }
 
     //presentations
@@ -522,6 +531,109 @@ final class OAuth2PresentationApiController extends OAuth2ProtectedController
                 SerializerUtils::getFields(),
                 SerializerUtils::getRelations()
             ));
+        });
+    }
+
+    #[OA\Put(
+        path: "/api/v1/summits/{id}/presentations/{presentation_id}/submission-period/reopen",
+        summary: "Admin-only: reopen the submission period for a presentation",
+        operationId: "reopenSubmissionPeriod",
+        security: [['summit_presentations_auth' => [SummitScopes::WriteSummitData, SummitScopes::WriteEventData, SummitScopes::WritePresentationData]]],
+        tags: ['Presentations'],
+        parameters: [
+            new OA\Parameter(name: 'id', in: 'path', required: true, schema: new OA\Schema(type: 'integer')),
+            new OA\Parameter(name: 'presentation_id', in: 'path', required: true, schema: new OA\Schema(type: 'integer')),
+        ],
+        requestBody: new OA\RequestBody(
+            required: false,
+            content: new OA\JsonContent(
+                properties: [
+                    new OA\Property(property: 'hours', type: 'integer'),
+                ]
+            )
+        ),
+        responses: [
+            new OA\Response(
+                response: Response::HTTP_CREATED,
+                description: "Created",
+                content: new OA\JsonContent(ref: "#/components/schemas/Presentation")
+            ),
+            new OA\Response(response: Response::HTTP_UNAUTHORIZED, description: "Unauthorized"),
+            new OA\Response(response: Response::HTTP_FORBIDDEN, description: "Forbidden"),
+            new OA\Response(response: Response::HTTP_NOT_FOUND, description: "Not Found"),
+            new OA\Response(response: Response::HTTP_PRECONDITION_FAILED, description: "Validation Error"),
+            new OA\Response(response: Response::HTTP_INTERNAL_SERVER_ERROR, description: "Server Error"),
+        ]
+    )]
+    public function reopenSubmissionPeriod($summit_id, $presentation_id)
+    {
+        return $this->processRequest(function () use ($summit_id, $presentation_id) {
+
+            $summit = SummitFinderStrategyFactory::build($this->summit_repository, $this->resource_server_context)->find($summit_id);
+            if (is_null($summit)) return $this->error404();
+
+            $current_member = $this->resource_server_context->getCurrentUser();
+            if (is_null($current_member)) return $this->error403();
+
+            $isAdmin = $current_member->isAdmin()
+                || $current_member->hasPermissionForOnGroup($summit, IGroup::SummitAdministrators);
+            if (!$isAdmin) return $this->error403();
+
+            $payload = $this->getJsonPayload(['hours' => 'sometimes|integer|min:1']);
+
+            // null, not the default: the hours rule (default AND ceiling) lives in the service.
+            $presentation = $this->presentation_submission_reopen_service->reopen(
+                $summit,
+                intval($presentation_id),
+                isset($payload['hours']) ? intval($payload['hours']) : null,
+                $current_member
+            );
+
+            // Private, NOT Admin: SerializerRegistry has no Admin key for Presentation and an
+            // unknown type silently falls back to Public, stripping the reopen fields.
+            return $this->updated(SerializerRegistry::getInstance()->getSerializer(
+                $presentation, SerializerRegistry::SerializerType_Private
+            )->serialize());
+        });
+    }
+
+    #[OA\Delete(
+        path: "/api/v1/summits/{id}/presentations/{presentation_id}/submission-period/reopen",
+        summary: "Admin-only: close the reopened submission period for a presentation",
+        operationId: "closeSubmissionPeriod",
+        security: [['summit_presentations_auth' => [SummitScopes::WriteSummitData, SummitScopes::WriteEventData, SummitScopes::WritePresentationData]]],
+        tags: ['Presentations'],
+        parameters: [
+            new OA\Parameter(name: 'id', in: 'path', required: true, schema: new OA\Schema(type: 'integer')),
+            new OA\Parameter(name: 'presentation_id', in: 'path', required: true, schema: new OA\Schema(type: 'integer')),
+        ],
+        responses: [
+            new OA\Response(response: Response::HTTP_NO_CONTENT, description: "No Content"),
+            new OA\Response(response: Response::HTTP_UNAUTHORIZED, description: "Unauthorized"),
+            new OA\Response(response: Response::HTTP_FORBIDDEN, description: "Forbidden"),
+            new OA\Response(response: Response::HTTP_NOT_FOUND, description: "Not Found"),
+            new OA\Response(response: Response::HTTP_INTERNAL_SERVER_ERROR, description: "Server Error"),
+        ]
+    )]
+    public function closeSubmissionPeriod($summit_id, $presentation_id)
+    {
+        return $this->processRequest(function () use ($summit_id, $presentation_id) {
+
+            $summit = SummitFinderStrategyFactory::build($this->summit_repository, $this->resource_server_context)->find($summit_id);
+            if (is_null($summit)) return $this->error404();
+
+            $current_member = $this->resource_server_context->getCurrentUser();
+            if (is_null($current_member)) return $this->error403();
+
+            $isAdmin = $current_member->isAdmin()
+                || $current_member->hasPermissionForOnGroup($summit, IGroup::SummitAdministrators);
+            if (!$isAdmin) return $this->error403();
+
+            $this->presentation_submission_reopen_service->closeNow(
+                $summit, intval($presentation_id), $current_member
+            );
+
+            return $this->deleted();
         });
     }
 
--- a/app/ModelSerializers/Summit/Presentation/AdminPresentationSerializer.php
+++ b/app/ModelSerializers/Summit/Presentation/AdminPresentationSerializer.php
@@ -11,6 +11,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  **/
+use Libs\ModelSerializers\One2ManyExpandSerializer;
 use models\summit\Presentation;
 
 /**
@@ -43,7 +44,27 @@ class AdminPresentationSerializer extends PresentationSerializer
         'TrackChairAvgScoresPerRakingType' => 'track_chair_scores_avg:json_string_array',
         'SubmissionReopenedUntil' => 'submission_reopened_until:datetime_epoch',
         'SubmissionReopenedById' => 'submission_reopened_by_id:json_int',
-        'SubmissionReopenedByNice' => 'submission_reopened_by:json_string',
+    ];
+
+    /**
+     * Declared HERE and not on the base PresentationSerializer on purpose. getExpandsMappings()
+     * merges parent into child only, and SubmissionPresentationSerializer is a SIBLING of this
+     * class (both extend PresentationSerializer), so this relation is unreachable from the
+     * Submission and Public variants. A case in the base expand switch would not be -- that is
+     * the leak the Admin-only design exists to prevent, and why the SDS rejected id+expand when
+     * a base-class switch was the only mechanism considered.
+     *
+     * serializer_type is explicit because One2ManyExpandSerializer defaults to Public, which
+     * blanks the actor's email.
+     */
+    protected static $expand_mappings = [
+        'submission_reopened_by' => [
+            'type' => One2ManyExpandSerializer::class,
+            'original_attribute' => 'submission_reopened_by_id',
+            'getter' => 'getSubmissionReopenedBy',
+            'has' => 'hasSubmissionReopenedBy',
+            'serializer_type' => SerializerRegistry::SerializerType_Private,
+        ],
     ];
 
     protected static $allowed_fields = [
@@ -70,6 +91,9 @@ class AdminPresentationSerializer extends PresentationSerializer
         'overflow_stream_key',
         'submission_reopened_until',
         'submission_reopened_by_id',
+    ];
+
+    protected static $allowed_relations = [
         'submission_reopened_by',
     ];
 
--- a/app/ModelSerializers/Summit/Presentation/AdminPresentationSerializer.php
+++ b/app/ModelSerializers/Summit/Presentation/AdminPresentationSerializer.php
@@ -41,6 +41,9 @@ class AdminPresentationSerializer extends PresentationSerializer
         'OverflowStreamIsSecure' => 'overflow_stream_is_secure:json_boolean',
         'OverflowStreamKey' => 'overflow_stream_key:json_string',
         'TrackChairAvgScoresPerRakingType' => 'track_chair_scores_avg:json_string_array',
+        'SubmissionReopenedUntil' => 'submission_reopened_until:datetime_epoch',
+        'SubmissionReopenedById' => 'submission_reopened_by_id:json_int',
+        'SubmissionReopenedByNice' => 'submission_reopened_by:json_string',
     ];
 
     protected static $allowed_fields = [
@@ -64,7 +67,10 @@ class AdminPresentationSerializer extends PresentationSerializer
         'etherpad_link',
         'overflow_streaming_url',
         'overflow_stream_is_secure',
-        'overflow_stream_key'
+        'overflow_stream_key',
+        'submission_reopened_until',
+        'submission_reopened_by_id',
+        'submission_reopened_by',
     ];
 
     /**
--- a/app/Models/Foundation/Summit/Events/Presentations/Presentation.php
+++ b/app/Models/Foundation/Summit/Events/Presentations/Presentation.php
@@ -49,6 +49,23 @@ class Presentation extends SummitEvent implements IPublishableEventWithSpeakerCo
 
     use OrderableChilds;
 
+    /**
+     * Redeclared, not extended: a child property SHADOWS the parent default, so SummitEvent's
+     * created_by/updated_by entries must be repeated here or getCreatedById()/getUpdatedById()
+     * stop resolving through One2ManyPropertyTrait::__call.
+     */
+    protected $getIdMappings = [
+        'getCreatedById' => 'created_by',
+        'getUpdatedById' => 'updated_by',
+        'getSubmissionReopenedById' => 'submission_reopened_by',
+    ];
+
+    protected $hasPropertyMappings = [
+        'hasCreatedBy' => 'created_by',
+        'hasUpdatedBy' => 'updated_by',
+        'hasSubmissionReopenedBy' => 'submission_reopened_by',
+    ];
+
     const ClassName = 'Presentation';
 
     const FieldProblemAddressed = 'problem_addressed';
@@ -2564,35 +2581,6 @@ SQL;
     public function getSubmissionReopenedBy(): ?Member
     {
         return $this->submission_reopened_by;
-    }
-
-    public function getSubmissionReopenedById(): int
-    {
-        try {
-            return is_null($this->submission_reopened_by) ? 0 : $this->submission_reopened_by->getId();
-        } catch (\Exception $ex) {
-            return 0;
-        }
-    }
-
-    /**
-     * Preformatted for Show Admin. Serialized as a plain string because an $array_mappings
-     * entry invokes scalar getters only and cannot serialize a Member.
-     *
-     * Yes, this departs from the repo's usual "who did this" idiom (a *_by_id scalar plus an
-     * expand case, e.g. CreatedById on SummitEventSerializer). That idiom was considered and
-     * REJECTED by the signed-off SDS, because the expand switch lives in the base
-     * PresentationSerializer, so a case added there is reachable from the Public and Submission
-     * variants too -- which is the exact leak the Admin-only design exists to prevent. The SDS
-     * mandates this shape ("no expand needed"), and Show Admin reads the field straight off the
-     * getEvent payload. If a relation is ever wanted, the safe mechanism is a subclass-local
-     * $expand_mappings, never a base-class switch case. Do not "correct" this to id+expand.
-     */
-    public function getSubmissionReopenedByNice(): ?string
-    {
-        $member = $this->submission_reopened_by;
-        if (is_null($member)) return null;
-        return sprintf("%s (%s)", $member->getFullName(), $member->getEmail());
     }
 
     /**
--- a/app/Models/Foundation/Summit/Events/Presentations/Presentation.php
+++ b/app/Models/Foundation/Summit/Events/Presentations/Presentation.php
@@ -232,6 +232,25 @@ class Presentation extends SummitEvent implements IPublishableEventWithSpeakerCo
     protected $custom_order;
 
     /**
+     * @var int|null the raw admin-granted duration; the window end is derived, never stored
+     */
+    #[ORM\Column(name: 'SubmissionReopenedHours', type: 'integer', nullable: true)]
+    protected $submission_reopened_hours = null;
+
+    /**
+     * @var \DateTime|null
+     */
+    #[ORM\Column(name: 'SubmissionReopenedDate', type: 'datetime', nullable: true)]
+    protected $submission_reopened_date = null;
+
+    /**
+     * @var Member|null
+     */
+    #[ORM\JoinColumn(name: 'SubmissionReopenedByID', referencedColumnName: 'ID', onDelete: 'SET NULL')]
+    #[ORM\ManyToOne(targetEntity: \models\main\Member::class, fetch: 'EXTRA_LAZY')]
+    protected $submission_reopened_by = null;
+
+    /**
      * @var PresentationSpeaker
      */
     #[ORM\JoinColumn(name: 'ModeratorID', referencedColumnName: 'ID', onDelete: 'SET NULL')]
@@ -363,6 +382,9 @@ class Presentation extends SummitEvent implements IPublishableEventWithSpeakerCo
         $this->attending_media = false;
         $this->will_all_speakers_attend = false;
         $this->disclaimer_accepted_date = null;
+        $this->submission_reopened_hours = null;
+        $this->submission_reopened_date = null;
+        $this->submission_reopened_by = null;
         $this->custom_order = 0;
         $this->track_chairs_scores = new ArrayCollection();
     }
@@ -2524,12 +2546,117 @@ SQL;
         return $review_status;
     }
 
+    public function getSubmissionReopenedHours(): ?int
+    {
+        return $this->submission_reopened_hours;
+    }
+
+    public function getSubmissionReopenedDate(): ?\DateTime
+    {
+        return $this->submission_reopened_date;
+    }
+
+    public function setSubmissionReopenedDate(?\DateTime $date): void
+    {
+        $this->submission_reopened_date = $date;
+    }
+
+    public function getSubmissionReopenedBy(): ?Member
+    {
+        return $this->submission_reopened_by;
+    }
+
+    public function getSubmissionReopenedById(): int
+    {
+        try {
+            return is_null($this->submission_reopened_by) ? 0 : $this->submission_reopened_by->getId();
+        } catch (\Exception $ex) {
+            return 0;
+        }
+    }
+
+    /**
+     * Preformatted for Show Admin. Serialized as a plain string because an $array_mappings
+     * entry invokes scalar getters only and cannot serialize a Member.
+     *
+     * Yes, this departs from the repo's usual "who did this" idiom (a *_by_id scalar plus an
+     * expand case, e.g. CreatedById on SummitEventSerializer). That idiom was considered and
+     * REJECTED by the signed-off SDS, because the expand switch lives in the base
+     * PresentationSerializer, so a case added there is reachable from the Public and Submission
+     * variants too -- which is the exact leak the Admin-only design exists to prevent. The SDS
+     * mandates this shape ("no expand needed"), and Show Admin reads the field straight off the
+     * getEvent payload. If a relation is ever wanted, the safe mechanism is a subclass-local
+     * $expand_mappings, never a base-class switch case. Do not "correct" this to id+expand.
+     */
+    public function getSubmissionReopenedByNice(): ?string
+    {
+        $member = $this->submission_reopened_by;
+        if (is_null($member)) return null;
+        return sprintf("%s (%s)", $member->getFullName(), $member->getEmail());
+    }
+
+    /**
+     * Derived, never stored — so hours and date cannot drift out of lockstep.
+     */
+    public function getSubmissionReopenedUntil(): ?\DateTime
+    {
+        if (is_null($this->submission_reopened_date) || is_null($this->submission_reopened_hours))
+            return null;
+        return (clone $this->submission_reopened_date)
+            ->add(new \DateInterval(sprintf("PT%dH", $this->submission_reopened_hours)));
+    }
+
+    /**
+     * A grant is only honored on a plan that is assigned, enabled, and whose submission
+     * window has actually ended. Deadline alone would grant edits before the CFP opened,
+     * and (via the isSubmissionClosed() fold) would open speaker deletes on a plan disabled
+     * after the grant — the delete path never re-checks IsEnabled() itself.
+     */
+    public function isSubmissionReopened(): bool
+    {
+        $selection_plan = $this->selection_plan;
+        if (is_null($selection_plan)) return false;
+        if (!$selection_plan->IsEnabled()) return false;
+
+        $submission_end_date = $selection_plan->getSubmissionEndDate();
+        if (is_null($submission_end_date)) return false;
+
+        $now = new \DateTime('now', new \DateTimeZone('UTC'));
+        // window not ended yet (pre-open or still open): there is nothing to reopen
+        if ($now <= $submission_end_date) return false;
+
+        $until = $this->getSubmissionReopenedUntil();
+        return !is_null($until) && $now < $until;
+    }
+
+    public function reopenSubmission(int $hours, Member $actor): void
+    {
+        $this->submission_reopened_hours = $hours;
+        $this->submission_reopened_date = new \DateTime('now', new \DateTimeZone('UTC'));
+        $this->submission_reopened_by = $actor;
+    }
+
+    /**
+     * Deliberately exempt from the invariants above: a stale grant must always be clearable.
+     */
+    public function closeSubmissionNow(): void
+    {
+        $this->submission_reopened_hours = null;
+        $this->submission_reopened_date = null;
+        $this->submission_reopened_by = null;
+    }
+
     /**
      * @return bool
      * @throws \Exception
      */
     public function isSubmissionClosed(): bool
     {
+        // Fold: covers both production callers (PresentationService::deletePresentation and
+        // SummitService) with no change at either call site. Review-status-safe because
+        // getReviewStatus() recomputes submission_closed inline rather than calling this.
+        if ($this->isSubmissionReopened()) return false;
+
         $selection_plan = $this->selection_plan;
         if (is_null($selection_plan)) return false;
 
--- a/app/Services/Model/Imp/PresentationSubmissionReopenService.php
+++ b/app/Services/Model/Imp/PresentationSubmissionReopenService.php
@@ -38,9 +38,16 @@ final class PresentationSubmissionReopenService
             if (!$presentation instanceof Presentation)
                 throw new EntityNotFoundException(sprintf("Presentation %s not found.", $presentation_id));
 
+            $max = intval(Config::get('cfp.max_reopen_hours', 168));
+
             // whole hours rule in one place: null means "unspecified", so the default is resolved
             // here rather than in the controller, right next to the ceiling it has to respect.
-            $hours = $hours ?? intval(Config::get('cfp.default_reopen_hours', 24));
+            // The resolved default is clamped to the ceiling on purpose: a deployment that
+            // configures default_reopen_hours above max_reopen_hours would otherwise reject every
+            // request that omits hours, which is a config error the caller cannot see or fix.
+            // An explicitly supplied $hours is still validated strictly below, never clamped.
+            if (is_null($hours))
+                $hours = min(intval(Config::get('cfp.default_reopen_hours', 24)), $max);
 
             // The lower bound is currently unreachable over HTTP -- the endpoint validates
             // 'hours' => 'sometimes|integer|min:1' and refuses first -- but it is deliberate, not
@@ -48,7 +55,6 @@ final class PresentationSubmissionReopenService
             // service), and a persisted non-positive value would make
             // Presentation::getSubmissionReopenedUntil() throw on every read, since
             // new \DateInterval('PT-1H') is invalid. Do not remove it as unused.
-            $max = intval(Config::get('cfp.max_reopen_hours', 168));
             if ($hours < 1 || $hours > $max)
                 throw new ValidationException(sprintf("hours must be between 1 and %s.", $max));
 
--- a/app/Services/Model/Imp/PresentationSubmissionReopenService.php
+++ b/app/Services/Model/Imp/PresentationSubmissionReopenService.php
@@ -83,29 +83,44 @@ final class PresentationSubmissionReopenService
 
     public function closeNow(Summit $summit, int $presentation_id, Member $actor): void
     {
-        $this->tx_service->transaction(function () use ($summit, $presentation_id, $actor) {
+        // closeSubmissionNow() nulls all three columns, the granting actor included, so the audit
+        // fields have to be read before the write and carried out of the closure. The line itself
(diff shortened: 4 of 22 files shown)


app/Http/Controllers/Apis/Protected/Summit/OAuth2PresentationApiController.php
```diff
@@ -525,6 +534,109 @@ public function updatePresentationSubmission($summit_id, $presentation_id)
...
+                    new OA\Property(property: 'hours', type: 'integer'),
+                ]
+            )
+        ),
+        responses: [
+            new OA\Response(
+                response: Response::HTTP_CREATED,
+                description: "Created",
+                content: new OA\JsonContent(ref: "#/components/schemas/Presentation")
+            ),
```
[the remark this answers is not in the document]
caseylocker: Refuted, and applying this would make the published spec wrong.

`JsonController::updated()` returns 201, not 200:

```php
protected function updated($data = 'ok', $has_content = true)
{
    $res = Response::json($data, $has_content ? 201 : 204);
```

`$has_content` defaults to true and this action does not override it, so the response is 201 and the existing `Response::HTTP_CREATED` declaration is correct. `testAdminCanReopenAClosedSubmission` asserts 201 and passes, which confirms it independently of reading the helper.

Leaving the attribute as is.

app/Services/ModelServicesProvider.php
```diff
@@ -193,6 +195,8 @@ public function register()
 
         App::singleton(IPresentationService::class, PresentationService::class);
 
+        App::singleton(IPresentationSubmissionReopenService::class, PresentationSubmissionReopenService::class);
+
```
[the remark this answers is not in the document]
caseylocker: Refuted: this provider is not deferred, so `provides()` is never consulted.

`protected $defer = true;` was the deferral mechanism in Laravel 5.7 and earlier. Since 5.8 a provider is deferred only if it implements `Illuminate\Contracts\Support\DeferrableProvider`, which this class does not. From the framework source in this repo's vendor tree:

```php
public function isDeferred()
{
    return $this instanceof DeferrableProvider;
}
```

This project runs laravel/framework 12.62.0, so `$defer` is dead metadata that the framework never reads. The provider loads eagerly and every binding in `register()` is available regardless of `provides()`.

Corroborating: `ISubmitterService` is registered in the same method and is also absent from `provides()`, with nothing failing in production.

The stale `$defer` property is arguably worth deleting, but that is pre-existing and out of scope for this PR.

tests/PresentationReopenApiTest.php
```diff
@@ -0,0 +1,701 @@
...
+        $this->reopen(['hours' => 24]);
+
+        $payload = SerializerRegistry::getInstance()->getSerializer(
+            $this->reloadPresentation(), SerializerRegistry::SerializerType_Public
+        )->serialize();
+
+        $this->assertArrayNotHasKey('submission_reopened_until', $payload);
+        $this->assertArrayNotHasKey('submission_reopened_by_id', $payload);
+        $this->assertArrayNotHasKey('submission_reopened_by', $payload);
+    }
```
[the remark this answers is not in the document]
caseylocker: Agreed, fixed in d673aa7db.

The test now asserts the reopen returned 201 and that the grant actually persisted before checking the three fields are absent:

```php
$this->reopen(['hours' => 24]);
$this->assertResponseStatus(201);
$reloaded = $this->reloadPresentation();
$this->assertTrue($reloaded->isSubmissionReopened(), 'grant did not persist');
```

You were right about the failure mode: a failed reopen leaves no grant, so Public would omit those fields whether or not the mappings are correctly scoped to the Admin subclass, and all three assertions would pass while proving nothing.

Only this test changed. As you noted, `testByFieldsAreAbsentFromTheSubmissionSerializer` is already covered by its `assertFutureEpoch` call on the same response.

caseylocker:
Dispositions for the five non-blocking items in the review body, since they are not inline threads. Two taken, three declined with reasons. The two inline bug reports are refuted in their own threads.

**Taken, in d673aa7db:**

1. **Pass the request serialization parameters.** The reopen response now forwards `expand`, `fields` and `relations` to `serialize()`, matching every sibling action in this controller. The `SerializerType_Private` argument is unchanged, so the Admin only fields stay Admin only.

2. **Guard against `default_reopen_hours` greater than `max_reopen_hours`.** A resolved default above the ceiling is now clamped to it rather than rejecting every request that omits `hours`. That is a configuration error the caller can neither see nor fix, so failing them is the wrong behaviour. An explicitly supplied `hours` is still validated strictly and never clamped, and a new unit test covers the clamp.

**Declined:**

3. **Duplicated reload helper, and `clear()` versus `refresh()`.** The duplication between the two test classes is deliberate. They need materially different identities, one global admin and one genuinely non admin, so a shared trait would need conditionals. On the mechanism: `clear()` is required rather than preferred. Each simulated HTTP dispatch closes the model entity manager, because `DoctrineMiddleware` closes it after every request and singleton repositories pin whichever manager instance they were first resolved with. A targeted `refresh()` cannot refresh an entity that is already detached. Teardown re-fetches its own roots, and the suite passes in both orders.

4. **`closeNow` accepts `$actor` but never uses it.** Intentional and documented on the interface. Closing nulls the granting actor column rather than restamping it, and the caller is already captured by generic request auditing. The signed off SDS specifies the parameter, so removing it would need an amendment.

5. **Rename or split the `PresentationMediaUploads` suite entry.** Accurate observation, no correctness impact. Deliberately left alone to keep this diff scoped.

The docstring coverage warning is not a PR specific correctness finding.

app/Models/Foundation/Summit/Events/Presentations/Presentation.php
```diff
@@ -2524,12 +2546,117 @@ public function getReviewStatusNice(): string
...
+    }
+
+    public function getSubmissionReopenedById(): int
+    {
+        try {
+            return is_null($this->submission_reopened_by) ? 0 : $this->submission_reopened_by->getId();
+        } catch (\Exception $ex) {
+            return 0;
+        }
+    }
```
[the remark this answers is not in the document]
caseylocker: Declining this one, for two reasons.

**It is the established idiom in this hierarchy, not a one off.** This accessor is a hand written copy of `One2ManyPropertyTrait::getPropertyId`, which is what every `get*ById()` accessor on these entities routes through:

```php
public function getPropertyId(string $property_name){
    try {
        return is_null($this->{$property_name}) ? 0 : $this->{$property_name}->getId();
    }
    catch(\Exception $ex){
        return 0;
    }
}
```

Same shape, same silent catch, same `0` sentinel. `SummitEvent` gets `getCreatedById` and `getUpdatedById` from exactly this through its `$getIdMappings` table, and there are around 14 similar accessors under `app/Models/`. Changing only this one would make it the odd one out.

**The suggested replacement is not equivalent.** `$this->submission_reopened_by?->getId() ?? 0` handles a null property, but the try/catch is there for a throwing proxy: this is an `EXTRA_LAZY` `ManyToOne`, and a proxy whose target row is gone raises `EntityNotFoundException` from `getId()`. Null safe access does nothing about that, so the change would turn a swallowed edge case into a 500 during serialization, which is the opposite of the intent.

Your underlying point about silent catches hurting debuggability is fair. It applies to the shared trait rather than to this PR, so it would be a repo wide change with its own risk profile, and it is out of scope here.

tests/PresentationReopenApiTest.php
```diff
@@ -0,0 +1,706 @@
...
+
+    // ---------------------------------------------------------------------------------------------
+    // Acceptance: an active grant actually lets the speaker edit after the window closed.
+    // The token member is both a global admin AND the presentation's creator with a speaker profile
+    // (InsertMemberTestData.php:144-149, plus setCreatedBy() in setUp above), so one token drives
+    // both roles. canEdit() matches on getCreatedById() == speaker->getMemberId()
+    // (Presentation.php:1264-1270).
+    // ---------------------------------------------------------------------------------------------
+
+    public function testActiveGrantLetsTheSpeakerUpdateAfterTheWindowClosed()
```
smarcet, line 567: @caseylocker The SDS's own acceptance bar — open-window parity (§8, D7) — has no test in this suite. §8 asks for three assertions under an active reopen grant: a field the selection plan marks non-editable stays non-editable, a field it marks editable stays editable, and the allowed media-upload types match the open-window set. The acceptance tests here prove the gate opens and closes correctly, but nothing pins the *editable surface*, so a future change that widens what a speaker can touch during a reopen window would pass this suite unnoticed.

Parity does hold today by construction — the diff only ORs the time gate, and `curatePayloadByPresentationAllowedQuestions` / `checkPresentationAllowedEdtiableQuestions` run unchanged on the update path (`PresentationService.php:557-558`) — but D7 calls this "the acceptance bar", which is exactly the kind of invariant that deserves a tripwire rather than an argument.

Suggested addition: two tests in this file — (1) grant a window, send an update touching a field the plan does not list as an allowed editable question, assert it is refused/stripped exactly as during the open window; (2) assert `getAllowedMediaUploadTypes()` under an active grant equals the open-window set. Note the scalar-field half of (1) is currently blocked by the pre-existing `SelectionPlan::areFieldsEqual` self-compare bug you already disclosed (`SelectionPlan.php:1570`) — that half belongs with that fix's ticket; the extra-questions and media-upload-type halves are writable now.

smarcet (changes requested):
@caseylocker please review

tests/PresentationReopenApiTest.php
caseylocker: Added in c91f721a9. Five tests.

Two of them are the real tripwires, and they use `links` rather than a scalar. `areFieldsEqual` has two branches: the array branch is correct, the scalar branch compares a value to itself, so the guard is dead for scalars. `links` is array typed in `AllowedEditableFields` and in `getSnapshot()`, so it takes the working branch and neither test needs the comparator fixed first.

- `testNonEditableFieldIsRefusedUnderReopen`: with `links` removed from the plan's allowed editable questions, an update under an active grant is refused with the plan's own message, and the link is not written.
- `testEditableFieldStaysEditableUnderReopen`: with it left in place, the same update returns 201 and the value persists.

You were right that the scalar half is blocked. It cannot assert enforcement at all, so it is covered as parity only, by `testNonEditableScalarHandledIdenticallyInTheOpenWindow` and `testNonEditableScalarHandledIdenticallyUnderReopen`. Both assert 201, which is the comparator defect showing through rather than a property worth keeping. When that comparator is fixed, both move to 412 together and both should be updated together. Only one moving is the signal the pair exists to give. Two tests rather than one because this suite gets one successful HTTP write per entity.

The media upload types assertion is in as `testAllowedMediaUploadTypesAreUnchangedUnderReopen`, with one caveat recorded in its docblock so a green is not over read. `getAllowedMediaUploadTypes()` returns the presentation type's stored collection, and `addMediaUploadTo` checks type, allowance and cap without consulting the submission window or the selection plan, so the set cannot vary with reopen state today. Media and speaker subresource mutations are currently ungated in every window, which is `86bba8388`. The assertion therefore holds by construction right now, and its value is as a guard against that changing rather than as evidence the gate is doing the work.

One fixture note for whoever reads these next: `SelectionPlan::__construct` seeds every allowed field as both a question and editable, so the restrictive case is built by removing a question, not by adding one.

Full file is green at 25 tests, 174 assertions. The whole `PresentationMediaUploads` shard is green at 59 tests, 385 assertions, and passed in CI on this commit.

caseylocker:
@smarcet Your request has been addressed. Ready for review

app/ModelSerializers/Summit/Presentation/AdminPresentationSerializer.php
```diff
@@ -64,7 +67,10 @@ class AdminPresentationSerializer extends PresentationSerializer
         'etherpad_link',
         'overflow_streaming_url',
         'overflow_stream_is_secure',
-        'overflow_stream_key'
+        'overflow_stream_key',
+        'submission_reopened_until',
+        'submission_reopened_by_id',
+        'submission_reopened_by',
```
smarcet, line 97: @caseylocker `submission_reopened_by` is a relation, and it should be declared as an expand mapping rather than flattened into a preformatted `"Full Name (email)"` string.

Reviewer note 4 says the SDS rejected the id+expand idiom because the expand switch lives in the base `PresentationSerializer`, so a case added there would be reachable from the Public and Submission variants. That concern is real — but it only applies to the base-class switch, and the fix is not to abandon expand. Your own docblock on `Presentation::getSubmissionReopenedByNice()` names the right mechanism: *"the safe mechanism is a subclass-local `$expand_mappings`, never a base-class switch case."* That is precisely the pattern in the file I linked: `SummitAttendeeSerializer.php:181` declares `protected static $expand_mappings` on the subclass, keyed per relation, with `original_attribute`, `getter`, `has` and an explicit `serializer_type`. Declared on `AdminPresentationSerializer` it is unreachable from the Public and Submission variants, so the leak the SDS was guarding against does not occur.

Why it matters beyond style: the flattened string is not consumable. A client that needs the actor's id, or wants to link to the member, has to parse a display string and guess at names containing parentheses. It also cannot be expanded, filtered, or reused, and it bakes a presentation decision into the model layer — `getSubmissionReopenedByNice()` exists only because the serializer could not express the relation.

Suggested fix: keep `submission_reopened_by_id:json_int` in `$array_mappings`, drop `SubmissionReopenedByNice` and the `submission_reopened_by` string field, and add to `AdminPresentationSerializer`:

```php
protected static $expand_mappings = [
    'submission_reopened_by' => [
        'type' => One2ManyExpandSerializer::class,
        'original_attribute' => 'submission_reopened_by_id',
        'getter' => 'getSubmissionReopenedBy',
        'has' => 'hasSubmissionReopenedBy',
        'serializer_type' => SerializerRegistry::SerializerType_Private,
    ],
];
```

Show Admin then reads `submission_reopened_by_id` off the `getEvent` payload as it does today, and gets the full member only when it asks. Worth noting the plumbing is already in place: `reopenSubmissionPeriod` forwards `SerializerUtils::getExpand()`, `getFields()` and `getRelations()` into `serialize()`, so `?expand=submission_reopened_by` would work the moment the mapping exists — the declaration is the only missing piece. If the SDS text needs an amendment to record this, that is worth doing — the constraint it encoded is satisfied by the subclass-local form.

app/Models/Foundation/Summit/Events/Presentations/Presentation.php
```diff
@@ -231,6 +231,25 @@ public static function getAllowedEditableFields(): array
...
+     */
+    #[ORM\Column(name: 'SubmissionReopenedDate', type: 'datetime', nullable: true)]
+    protected $submission_reopened_date = null;
+
+    /**
+     * @var Member|null
+     */
+    #[ORM\JoinColumn(name: 'SubmissionReopenedByID', referencedColumnName: 'ID', onDelete: 'SET NULL')]
+    #[ORM\ManyToOne(targetEntity: \models\main\Member::class, fetch: 'EXTRA_LAZY')]
+    protected $submission_reopened_by = null;
```
smarcet, line 268: @caseylocker `getSubmissionReopenedById()` reimplements `One2ManyPropertyTrait::getPropertyId()` line for line — same null check, same `getId()`, same `catch (\Exception) { return 0; }`. The trait is already in play here: `SummitEvent` uses it (`SummitEvent.php:64`) and declares `$getIdMappings` for `created_by` / `updated_by`, so `Presentation` inherits the mechanism and hand-writing the getter opts out of it for no gain.

Why it matters: the trait is the single place where "id of a nullable relation" semantics live. A hand-rolled copy stops tracking it — if the trait's null/exception behaviour is ever changed (0 vs null, logging the swallowed exception), every declared mapping follows and this one does not.

Suggested fix — declare the mapping instead of the method, and **carry the parent's entries forward**, since a redeclared `$getIdMappings` on the subclass shadows `SummitEvent`'s rather than merging with it, which would silently kill `getCreatedById()` / `getUpdatedById()`:

```php
protected $getIdMappings = [
    'getCreatedById' => 'created_by',
    'getUpdatedById' => 'updated_by',
    'getSubmissionReopenedById' => 'submission_reopened_by',
];

protected $hasPropertyMappings = [
    'hasCreatedBy' => 'created_by',
    'hasUpdatedBy' => 'updated_by',
    'hasSubmissionReopenedBy' => 'submission_reopened_by',
];
```

The `hasSubmissionReopenedBy` half is what the expand mapping in my serializer comment needs. Your existing unit tests (`PresentationSubmissionReopenTest.php`, the `getSubmissionReopenedById` cases) cover both the 0 and the actor-id paths and should pass unchanged — worth confirming, since `__call()` only fires for methods that do not exist, so the hand-written one must actually be deleted.

config/cfp.php
```diff
@@ -17,4 +17,8 @@
     'support_email' => env('CFP_SUPPORT_EMAIL', null),
     'client_id' => env('CFP_OAUTH2_CLIENT_ID', null),
     'scopes' => env('CFP_OAUTH2_SCOPES', null),
-];
\ No newline at end of file
+
+    // ceiling on an admin-granted per-presentation reopen window
```
smarcet, line 21: @caseylocker The PR's Deploy notes cover the two migrations but not these two new config keys, and they are the ones an operator will get wrong. `CFP_MAX_REOPEN_HOURS` and `CFP_DEFAULT_REOPEN_HOURS` are read through `env()`, so on a deployed instance they resolve from the environment, not from `.env.example` — an unset variable falls back to 168 and 24, which happens to be the intended default, so a misconfiguration is invisible until an admin sees a window they did not expect.

Please add to the Deploy notes: the two variable names, their defaults and units (hours), which environments need them set explicitly versus left to default, and — given reviewer note 7 — that configuring `CFP_DEFAULT_REOPEN_HOURS` above `CFP_MAX_REOPEN_HOURS` is silently clamped rather than rejected, so a typo there produces no error anywhere. Also confirm whether the k8s configmap / secret for this service needs the keys added, since that is a separate change from `.env.example`.

database/migrations/config/Version20260807130000.php
```diff
@@ -0,0 +1,76 @@
...
+    use APIEndpointsMigrationHelper;
+
+    private const API_NAME = 'summits';
+
+    private const ENDPOINTS = [
+        [
+            'name' => 'reopen-presentation-submission-period',
+            'route' => '/api/v1/summits/{id}/presentations/{presentation_id}/submission-period/reopen',
+            'http_method' => 'PUT',
+            'scopes' => [
```
smarcet, line 51: @caseylocker Both entries need `authz_groups`. The migration's own docblock states "No authz_groups: the routes do not use `auth.user`" — that reasoning is inverted, see my comment on `routes/api_v1.php:867`: the empty group list is what forces the middleware out, so filling it is what lets the middleware back in. Fix both endpoint definitions here in the same change as the route and the seeder, and update the docblock so the next reader does not inherit the original rationale.

database/migrations/model/Version20260807120000.php
```diff
@@ -0,0 +1,73 @@
...
+{
+    public function getDescription(): string
+    {
+        return 'Add per-presentation CFP reopen override columns to Presentation.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql(<<<SQL
+            ALTER TABLE `Presentation`
```
smarcet: @caseylocker This migration should go through `LaravelDoctrine\Migrations\Schema\Builder` rather than raw `addSql()` throughout.

The docblock justifies the raw form by the ordering hazard: mixing `Builder` schema-diff with `addSql()` in one migration runs the `addSql()` statements first and the diff last, so the FK would execute before its column exists. That hazard is real — `Version20260615000001` documents exactly the same ordering in its own header, and relies on it deliberately. But the repo's answer to it is not to abandon `Builder`; it is to split the migration. `Version20260615000000` adds the column and its index through `Builder`:

```php
$builder->table(self::AddOnTable, function (Table $table) {
    $table->integer('AddOnTypeID', false, false)->setNotnull(false);
    $table->index('AddOnTypeID', 'IDX_SummitSponsorshipAddOn_AddOnTypeID');
});
```

and `Version20260615000001` adds the FK constraint separately, as the one statement that genuinely needs raw DDL. The ordering problem never arises because the two never share a migration.

Why it matters: raw DDL bypasses the platform abstraction and the `hasTable()` / `hasColumn()` guards the rest of the model migrations use, which is what makes them safe to re-run. That is directly relevant here — deploy note 2 warns that a partial failure leaves the columns in place while the migration is unrecorded, so a naive retry fails on duplicate objects and an operator has to inspect the schema by hand. With `Builder`'s existence guards, the retry is a no-op instead and the warning becomes unnecessary.

Suggested fix: split into two migrations, matching `Version2026061500000{0,1}`:

1. the three columns plus the index on `SubmissionReopenedByID`, through `$builder->table('Presentation', ...)` with a `hasColumn()` guard;
2. the FK constraint as `addSql()`, in its own migration.

`down()` should mirror that split. If a single migration is required for some reason, then at minimum guard the column adds so a partial failure is recoverable, and say so in the docblock.

database/seeders/ApiEndpointsSeeder.php
```diff
@@ -7090,6 +7090,27 @@ private function seedSummitEndpoints()
                     SummitScopes::WritePresentationSlidesData
                 ],
             ],
+            // submission period reopen (admin-only; authorization enforced in the controller)
+            [
+                'name' => 'reopen-presentation-submission-period',
+                'route' => '/api/v1/summits/{id}/presentations/{presentation_id}/submission-period/reopen',
+                'http_method' => 'PUT',
+                'scopes' => [
```
smarcet, line 7099: @caseylocker Same as the config migration: both entries need `authz_groups`, and the inline comment "admin-only; authorization enforced in the controller" should go with them. Full reasoning on `routes/api_v1.php:867`. Keep the seeder and `Version20260807130000.php` byte-identical on the group list — they drift silently otherwise, and a fresh install then behaves differently from a deployed one.

routes/api_v1.php
```diff
@@ -860,6 +860,15 @@
                     });
                 });
 
+                // submission period (admin-only reopen; authorization is done in the controller,
+                // NOT via auth.user -- that middleware 403s any endpoint with no authz groups)
+                Route::group(['prefix' => 'submission-period'], function () {
+                    Route::group(['prefix' => 'reopen'], function () {
+                        Route::put('', 'OAuth2PresentationApiController@reopenSubmissionPeriod');
```
smarcet: @caseylocker Reviewer note 1 is factually right about the mechanism and wrong about the conclusion. `UserAuthEndpoint::handle()` does 403 when the endpoint's group list is empty (`app/Http/Middleware/UserAuthEndpoint.php:129-149` — the loop finds no match and falls through to the 403), but that is a consequence of registering the endpoint with zero `authz_groups`, not an independent reason to drop `auth.user`. Add the groups and the failure mode disappears.

Why it matters: as written, the only authorization on these two routes is the summit-aware admin check inside the controller. That check runs after routing and after the controller is resolved, so it is one refactor, one inherited method, or one copy-pasted action away from being silently absent — and nothing in the route definition would show it. Every other admin-mutating presentation route in this group goes through the middleware, which fails closed at the routing layer regardless of what the controller does. Note also that `handle()` returns early at line 73 when there is no current member, so `auth.user` does not break pure service-token callers.

The media-upload routes cited as precedent are a known exception, not the house pattern; reproducing an exception widens it.

Suggested fix, one change across three files:

1. here and at line 868, add `'middleware' => 'auth.user'`;
2. add the appropriate `authz_groups` to both entries in `database/migrations/config/Version20260807130000.php`;
3. same for both entries in `database/seeders/ApiEndpointsSeeder.php:7094-7112`.

Use the same group set as the sibling presentation write endpoints (summit admins / super admins). If there is a specific group that cannot be assigned here, name it and I will take another look — but "no groups" is not a viable end state for an admin-only mutating endpoint.

smarcet (changes requested):
@caseylocker please re review

app/Services/Model/Imp/PresentationSubmissionReopenService.php
```diff
@@ -0,0 +1,95 @@
...
+            if ($now <= $submission_end_date)
+                throw new ValidationException("Submission period has not ended yet; nothing to reopen.");
+
+            $presentation->reopenSubmission($hours, $actor);
+
+            return $presentation;
+        });
+    }
+
+    public function closeNow(Summit $summit, int $presentation_id, Member $actor): void
```
smarcet, line 84: @caseylocker Backing CodeRabbit's nitpick here, and I'd rate it above trivial. `$actor` is not just unused — it never enters the closure's `use` list, so the parameter is inert by construction rather than by an oversight in one branch.

The reason it matters is what `closeSubmissionNow()` does: it nulls all three columns, `submission_reopened_by` included. So closing a grant erases who granted it and records nothing about who revoked it. Both ends of an admin-only privilege operation end up unauditable, which sits badly against the migration's own `down()` docblock describing these columns as "the durable actor/date audit trail".

Two ways out, either is fine:

- log the revocation (`Log::info` with summit id, presentation id, actor id — cheap, keeps the schema as is); or
- drop `$actor` from `IPresentationSubmissionReopenService::closeNow()` and from this implementation, so the signature stops implying an audit that does not happen.

What should not stay is a parameter that documents an intent the code does not honour.

app/ModelSerializers/Summit/Presentation/AdminPresentationSerializer.php
caseylocker: Done in c1cffca1. `submission_reopened_by` is a relation now: the payload carries `submission_reopened_by_id`, and `?expand=submission_reopened_by` returns the serialized member. The preformatted string is gone.

One deliberate detail. The mapping is declared on `AdminPresentationSerializer`, not on the base. `AbstractSerializer::getExpandsMappings()` merges the class lineage parent to child only, and `SubmissionPresentationSerializer` is a sibling of this class, so the relation stays unreachable from the Submission and Public variants. That was the SDS's objection to id plus expand, and it does not apply to a subclass local mapping. `serializer_type` is Private, matching how `created_by` is already serialized here; the default is Public, which blanks the actor's email.

Covered by `testReopenActorIsExpandableOnTheAdminResponse`, which asserts the id key is replaced by the relation and that the relation carries the email.

On your last point, the SDS amendment: written and open as fntechgit/ftn-docsnsklz#115. It records the shape change, why the subclass local form satisfies the constraint §4 encoded, and the two places the body is now wrong about the shipped contract (§6's "no `getEvent` expand entry is needed" and §8's "plain scalars, no expand"). The body itself is untouched, the entry is additive. Vault `main` needs one non-author approval, so it is waiting on a reviewer.

app/Models/Foundation/Summit/Events/Presentations/Presentation.php
caseylocker: Done in c1cffca1. The hand written accessor is gone and `getSubmissionReopenedById()` resolves through `One2ManyPropertyTrait`, which `Presentation` inherits from `SummitEvent`. `hasSubmissionReopenedBy()` comes along with it and is what the new expand mapping uses for its `has` check.

One trap worth recording for the next person. A child property shadows the parent default, so `Presentation::$getIdMappings` has to repeat SummitEvent's `created_by` and `updated_by` entries. `__call` returns null for an unmapped name rather than raising, so omitting them would have made `getCreatedById()` return null silently instead of failing. `testInheritedCreatedByAccessorsStillResolve` pins that.

config/cfp.php
caseylocker: Added to the PR description as deploy notes 5 and 6.

Note 5 covers both values: `CFP_MAX_REOPEN_HOURS` (default 168, seven days) and `CFP_DEFAULT_REOPEN_HOURS` (default 24). Both are documented in `.env.example` and both defaults live in `config/cfp.php`, so no `.env` change is required to deploy; set them per environment only if a show needs a different ceiling.

Note 6 covers migration ordering, which the `auth.user` change makes load bearing: both migrations have to run with or before the application, the model one because the entity maps three columns only it creates, the config one because the routes now read their authz groups from it.

database/migrations/config/Version20260807130000.php
caseylocker: Added in 38d2714d: `SuperAdmins`, `Administrators` and `SummitAdministrators`, mirroring `update-event` minus the track chair roles, which have no business reopening a submission window.

`registerEndpoints()` already consumed an `authz_groups` key, so this is a data only change, and rollback is already covered since `api_endpoints` cascades to `endpoint_api_authz_groups`. The seeder carries the identical list.

database/seeders/ApiEndpointsSeeder.php
caseylocker: Added in 38d2714d, the same three groups as the config migration, kept in parity with it deliberately: the seeder covers fresh installs and the migration covers deployed environments, since the deploy flow does not re-run seeders.

routes/api_v1.php
caseylocker: Added in 38d2714d, on both routes, together with the authz groups the middleware reads.

Those two asks go together. `auth.user` returns 403 for any endpoint whose required group list is empty, which is why the middleware was left off originally; registering the groups is the right fix rather than skipping the middleware. It narrows nothing, because `Member::hasPermissionForOnGroup()` already ends in `&& isOnGroup($slug)`, so every caller that passes the controller's summit scoped admin check passes the middleware's global group check too. A client credentials token with no member passes the middleware and is still refused by the controller.

`PresentationReopenAuthzTest` is still green at 12 tests.

database/migrations/model/Version20260807120000.php
caseylocker: Reworked to the split in 8e51951e, matching `Version2026061500000{0,1}` as you laid out.

- `Version20260807120000`: the three nullable columns plus the index on `SubmissionReopenedByID`, entirely through `Builder`.
- `Version20260807120001`: the foreign key as its single `addSql()` statement. It only reads the schema, never mutates it, so it produces no diff and the ordering hazard cannot reappear inside it.
- `down()` mirrors the split, and the version order matters: `120001` drops the constraint first, which is what lets `120000` drop the index it sits on.

Your point about the guards is what the split made obvious. Both migrations now check each component on itself, so an interrupted run repairs on retry rather than failing on duplicate objects, and deploy note 2 has been rewritten accordingly. Driving the pair against synthetic schemas: fresh emits the columns and index then the constraint; columns-only emits the index and the constraint alone; a fully applied schema emits nothing from either.

One improvement that fell out of splitting: it restores the original statement order, index before constraint, so MySQL builds the constraint on the existing index instead of creating its own. Verified against a scratch database, one index named `SubmissionReopenedByID`, column types unchanged.

app/Services/Model/Imp/PresentationSubmissionReopenService.php
caseylocker: Agreed, and the diagnosis was exactly right: `$actor` never entered the closure's `use` list, so it was inert by construction rather than by a missed branch. Took the logging option, in 1f0346d8 and d8572c54.

The line is emitted **after** `transaction()` returns, not inside the closure. That mattered more than it first looked: `DoctrineTransactionService::transaction()` runs the callback and only then flushes and commits, and it re-runs the whole callback on a retryable failure. Logging inside would have announced a completed revocation that had not happened yet, and could have announced it more than once for a single request. The audit fields are still read before `closeSubmissionNow()` nulls them, then carried out of the closure. A transaction that throws logs nothing.

Two details in the line itself worth naming, since both were wrong in my first cut:

- `getSubmissionReopenedById()` returns `0` rather than null for an absent relation, and closing with no active grant is a supported no-op, so the line read "granted by member 0". It now reads "no active grant".
- The window end is normalized back to UTC and stamped with a `Z`. The value is written UTC but Doctrine hydrates datetimes in the application default timezone, so formatting it raw was one config change away from being wrong.

The signature is unchanged, so the audit the parameter implies is now the audit the code performs. Three tests cover it: `testCloseNowLogsTheRevocationWithBothActors`, `testCloseNowWithNoGrantLogsNoActiveGrantRatherThanMemberZero`, and `testCloseNowDoesNotLogWhenTheTransactionFails`.

config/cfp.php
caseylocker: Following up on your last question, the k8s side. I went and looked rather than guessing, and the answer is no change is needed, on either path.

The declaration site is `argocd-apps/summit-api/values-{prod,stage}.yaml`, under the `config:` block, which `templates/deployment.yaml` mounts onto the containers with `envFrom: configMapRef: <fullname>-config`. That is where the three existing keys live:

```yaml
    CFP_APP_BASE_URL: https://speakermgmt.fnvirtual.app
    CFP_SUPPORT_EMAIL: support@fntech.com
    # CFP_OAUTH2_CLIENT_ID: secret
```

Neither `CFP_MAX_REOPEN_HOURS` nor `CFP_DEFAULT_REOPEN_HOURS` appears anywhere in that repo on `origin/main`, in `values.yaml`, `values-prod.yaml` or `values-stage.yaml`.

You asked about the configmap **or the secret**, so I checked the secret side too rather than only the values files: the Doppler secret this chart's `ExternalSecret` syncs into `<fullname>-secrets`, in both the prod and stage configs. Neither key is present there either. That check mattered rather than being belt-and-braces, because `envFrom` lists `configMapRef` before `secretRef` on every php container, so a key set on the secret side would silently win over the values file.

So both environments take the compiled fallbacks of 168 and 24, which are the intended values, and this ships without touching the chart or the secret. They only need adding where a show wants a different ceiling, and since they are plain non-secret integers they belong inline in the `config:` block like `CFP_APP_BASE_URL`, not in the commented `: secret` form that routes through `externalsecrets.yaml`.

One timing constraint if they ever are added, which is not obvious and is the same invisible-misconfiguration shape you flagged for the clamp: `fn-docker/summit-api/php-entry-point.sh` runs `php artisan config:cache` on every container start, for every command. `config/cfp.php` resolves `env()` at cache build time, so the value has to be in the pod environment before the entrypoint runs. Anything injected after that is baked out and silently reads back as the default. That is now in deploy note 5 along with the names, defaults, units and the clamp.

The one thing I have not read is the parent chart that renders `<fullname>-config`, since it is a dependency rather than part of that repo. It does not change the answer: the consumption side is confirmed in `deployment.yaml`, and the three existing `CFP_*` keys prove the path works in prod.

database/migrations/model/Version20260807120001.php
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public function up(Schema $schema): void
+    {
+        $table = $schema->getTable(self::TableName);
+
+        // the column is Version20260807120000's job; if it is absent that migration has not run,
+        // and adding the constraint here would fail on a missing column rather than explain itself
+        if (!$table->hasColumn(self::ActorColumn)) return;
+        if ($table->hasForeignKey(self::FkName)) return;
```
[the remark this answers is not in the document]
caseylocker: Valid, and fixed in b3410a76.

Confirmed the mechanism before changing anything: `DbalExecutor` catches `SkipMigration` and calls `setSkipped(true)` without rethrowing, and treats any other `Throwable` as an error that it rethrows. A plain `return` is neither, so it is the ordinary success path and Doctrine records the version. Your consequence follows exactly: the version is stamped applied, the constraint is never created, and the column is left with no `ON DELETE SET NULL` and no migration remaining to add it.

One correction to the suggested fix, because it matters to anyone applying this pattern later. The right form is `AbstractMigration::abortIf()`, which throws `AbortMigration` and lands in that generic `Throwable` branch, so the run stops and the version stays unrecorded and re-runnable once the prerequisite is repaired. `skipIf()` would look like the natural choice for a guard and would reproduce the exact bug you reported, since it marks the version skipped and records it.

The `hasForeignKey()` early return is deliberately left as a return: an idempotent no-op is the correct outcome there, and recording the version is right.

Verified against the three states rather than by inspection:

- column missing: aborts, with a message naming `Version20260807120000` as the prerequisite
- columns present, no constraint: emits one `ADD CONSTRAINT`
- constraint already present: emits nothing

romanetar:
### Code review

Found 1 issue:

1. The `closeNow()` interface docblock still documents `$actor` as unused, but the implementation now uses it. This is the stale half of the fix for [smarcet's review comment](https://github.com/OpenStackweb/summit-api/pull/581#discussion_r3752951196) ("What should not stay is a parameter that documents an audit that does not happen") — the implementation was updated in `1f0346d81`, the interface contract was not. As written it invites a future reader to drop `$actor` and silently kill the revocation audit line.

https://github.com/OpenStackweb/summit-api/blob/b3410a76225afda7799c5a9aee407361bcee21fd/app/Services/Model/IPresentationSubmissionReopenService.php#L41-L45

The implementation does use it, at `$actor->getId()`:

https://github.com/OpenStackweb/summit-api/blob/b3410a76225afda7799c5a9aee407361bcee21fd/app/Services/Model/Imp/PresentationSubmissionReopenService.php#L115-L124

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

romanetar (changes requested):
@caseylocker please review

caseylocker:
Confirmed and fixed in `5a05eda04`.

You're right that the interface contract was the stale half. `1f0346d81` made `$actor` the subject of the revocation audit line but left the docblock calling it unused and justifying that on the grounds that the caller is already captured by generic request auditing, which is now the opposite of what the code does.

One thing worth adding to your read: the stale text is a bit worse than out of date. The "do not tidy it away" warning still lands on the right conclusion, but it gets there through a premise a reader can check and find false, which invites discounting the warning and dropping the parameter anyway. So I replaced the whole rationale rather than just the word "unused":

```
 * $actor is logged as the revoking member in the audit line this method emits. Removing it
 * silently drops that attribution from the revocation record.
```

Comment only, no behavior change.

caseylocker:
@romanetar @smarcet ready for review.

smarcet (approved):
LGTM

PR #585, opened 2026-08-14
fix(speakers): validate photo override exists before clearing

smarcet:
ref:
https://app.clickup.com/t/9014802374/86bb6hm53
## Summary
- `SpeakerService::clearSpeakerPhoto` / `clearSpeakerBigPhoto` now throw `ValidationException` when the speaker has no photo override, instead of silently no-oping.
- `PresentationSpeaker::getProfilePhotoUrl` / `getBigProfilePhotoUrl` no longer fall back to a Gravatar URL; removed the now-unused `getGravatarUrl()` helper as a result.
- Trailing-whitespace cleanup on unrelated DQL/SQL query strings in the same file.

## Test plan
- `vendor/bin/phpunit tests/Unit/Entities/PresentationSpeakerTest.php tests/SpeakerServiceTest.php` — same 4 pre-existing failures with and without this diff (stale DB schema / test fixtures, unrelated to this change).
- `php -l` on the modified model file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved speaker profile photo handling by removing unexpected Gravatar fallbacks.
  * Speaker photos now correctly use the resolved photo, configured default image, or no image when unavailable.
  * Added validation to prevent deletion of speaker photos when no corresponding override exists.
  * Standardized the error message shown when speaker photo deletion cannot be completed.

### DIFF
--- a/app/Models/Foundation/Summit/Speakers/PresentationSpeaker.php
+++ b/app/Models/Foundation/Summit/Speakers/PresentationSpeaker.php
@@ -746,7 +746,7 @@ class PresentationSpeaker extends SilverstripeBaseModel
         $private_tracks = [];
 
         if ($exclude_privates_tracks) {
-            $private_track_groups = $this->createQuery("SELECT pg from models\summit\PrivatePresentationCategoryGroup pg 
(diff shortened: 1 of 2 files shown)


romanetar (approved):
LGTM

